PR #58, opened 2019-01-24
rochaa/sso

xandroc:


chameleon/views.py
```diff
@@ -1,17 +1,106 @@
...
+            + request.user.username + ', callback host was: ' + host)
+        return HttpResponseRedirect('/')
+
+    ## get the Keystone token, and add to context 
+    ## if none available, prompt user for username/password to get one
+    context = {}
+    try:
+        context['sso_token'] = request.session['unscoped_token'].get('auth_token')
+    except:
+        return manual_ks_login(request)
```
diurnalist: Kind of not ideal that we have to maintain 2 separate login pages now, but I get why... the login form supports other methods of login, like GENI. Actually... I wonder how being logged in via GENI or something would affect this flow. (Do people login with GENI in practice?)

chameleon/views.py
```diff
@@ -1,17 +1,106 @@
...
+    return render(request, 'sso/manual_ks_login.html', context)
+
+## This page is meant for users who were trying to sso to Horizon, 
+## but were unable to retrieve a Keystone token
+@login_required
+def horizon_sso_unavailable(request):
+    try:
+        ## Here we're just checking to see if a keystone token is in the session,
+        ## if it is, we shouldn't be here, let's redirect to the home page
+        request.session['unscoped_token']
```
diurnalist: Instead of relying on a `KeyError` here, why not just do:

```python
if getattr(request.session, 'unscoped_token'):
    return redirect('/')
else:
    return render(request, 'sso/keystone_login_unavailable.html', None)
```

Code that catches `KeyError` feels odd to me in general, given that it's often not clear _where_ the KeyError came from, so then what is the code supposed to do?

chameleon/views.py
```diff
@@ -1,17 +1,106 @@
...
+    context['host'] = protocol + '://' + host + horizon_webroot + '/auth/ccwebsso/' + '?next=' + next
+    return render(request, 'sso/sso_callback_template.html', context)
+
+@login_required
+def manual_ks_login(request):
+    if request.method == 'POST':
+        form = KSAuthForm(request, data=request.POST)
+        user = None
+        if request.user.username == request.POST.get('username') and form.is_valid():
+            login.set_unscoped_token(request)
```
diurnalist: Having to check that `set_unscoped_token` _actually set the token_ is one of the consequences of having side-effects in functions like that. A more explicit way can get around this problem:

```python
# Change 'set' to 'get'
unscoped_token = login.get_unscoped_token(request)
if unscoped_token:
    # We know login was successful!
    request.session['unscoped_token'] = unscoped_token
    return horizon_sso_login(request)
else:
    return HttpResponseRedirect('/sso/horizon/unavailable')
```

In this case extracting `get_unscoped_token` to some sort of authentication helper would also make sense. Another refactoring step might be to have a helper function to set it on the request, but this would just be so that you don't have to remember what key on the session you're using:

```python
def set_session_unscoped_token(request, token):
    request.session['unscoped_token'] = token
```

chameleon/os_login.py
```diff
@@ -13,18 +14,29 @@
...
                 user_domain_name='default', user_domain_id='default')
             request.session['unscoped_token'] = unscoped_token
-            logger.info('*********************################## Unscoped Token retrieved and stored in session')
-        except Exception as err:
-            logger.error('Error retrieving Openstack Token: {}'.format(err.message) + str(sys.exc_info()[0]))
-    return login_return
+            logger.info('***Unscoped Token retrieved from OPENSTACK_ALT_KEYSTONE_URL and stored in session for user: ' + request.POST.get('username'))
+        except Exception as err2:
+            logger.error('Error retrieving Openstack Token from OPENSTACK_ALT_KEYSTONE_URL: {}'.format(err2.message) + str(sys.exc_info()[0]))
+    except Exception as err:
```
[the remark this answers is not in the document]
xandroc: I think you're right, looks like it was left over from some refactoring that function

chameleon/views.py
xandroc: Agree, this won't work for login, it doesn't create a django session and to get here we actually need to be logged in. The form allows users who either logged in with Geni or who's account was in a transient state at original login to retrieve a Keystone token. Both are edge cases, but since the portal has long sessions and users may not log out and back in after creating their portal account, this gives them a way to either get a token or get an informational message about why they can't log in to Horizon.

PR #67, opened 2019-09-12
Sharing portal integration

diurnalist:


chameleon/urls.py
```diff
@@ -76,12 +78,4 @@
...
-]
-
-if settings.DEBUG:
-    urlpatterns = [
-        url(r'^media/(?P<path>.*)$', serve,  # NOQA
-            {'document_root': settings.MEDIA_ROOT, 'show_indexes': True}),
-    ] + staticfiles_urlpatterns() + urlpatterns  # NOQA
-else:
-    urlpatterns = staticfiles_urlpatterns() + urlpatterns
+] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
diurnalist: This is the new way to do this: https://docs.djangoproject.com/en/1.11/howto/static-files/#serving-files-uploaded-by-a-user-during-development

PR #71, opened 2019-09-27
Ensuring no redirect loop on form post and Horizon SSO

xandroc:
The login loop may have surfaced when a token was retrieved successfully at the manual login form, but an error occurred when synchronizing projects. The error sent the user to `manual_ks_login(request)` where, if the original request method was POST, the function then interpreted the request as a form submission and tried to get a new token, then looped back to attempt project sync where the same failure as before caused another call to `manual_ks_login(request)`, causing a loop in Django

### DIFF
--- a/chameleon/views.py
+++ b/chameleon/views.py
@@ -32,11 +32,10 @@ WHITELISTED_PROJECTS = ['openstack', 'maintenance']
 def horizon_sso_login(request):
     unscoped_token = request.session.get('unscoped_token')
     '''
-    Because some users with previously active projects may be able to get auth tokens even if they can't access Horizon,
-    we check with Keystone to verify an active project exists
-    If user has an active project, we proceed to log them in
+    First, we check with Keystone to verify an active project exists (that process includes synchronizing TAS & Keystone projects)
+    If a user has a token and active project, we send the user to Horizon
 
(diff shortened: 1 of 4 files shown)


chameleon/views.py
```diff
@@ -48,38 +47,46 @@ def horizon_sso_login(request):
...
+            if chameleon_projects.get('approved_projects'):
+                earliest_start_date = None
+                for p in chameleon_projects.get('approved_projects'):
+                    for a in p.allocations:
+                        curr_alloc_date = datetime.strptime(a.start,'%Y-%m-%dT%H:%M:%SZ')
+                        if earliest_start_date is None or curr_alloc_date < earliest_start_date:
+                            earliest_start_date = curr_alloc_date
+                # send users to a page that lets them know when their approved project begins
+                logger.debug('User: ' + request.user.username + ' is attempting to log in to Horizon, approved project found in TAS but is not yet active, sending to informational page')
+                return render(request, 'sso/chameleon_project_approved.html', {'active_date': datetime.strftime(earliest_start_date,'%B %d, %Y at %I:%M %p'),})
```
diurnalist: What does this do if `earliest_start_date` is `None`, which can happen if there are no allocations?

PR #72, opened 2019-09-27
google analytics measurement protocol

zhenz:
Add a hook to send messages to GA for tracking the downloads of complex appliances.

### DIFF
--- a/appliance_catalog/views.py
+++ b/appliance_catalog/views.py
@@ -25,6 +25,8 @@ from keystoneauth1 import session
 from keystoneclient.v3 import client
 from glanceclient import Client
 from smtplib import SMTPException
+from six.moves import http_client, urllib
+import uuid
 
 logger = logging.getLogger('default')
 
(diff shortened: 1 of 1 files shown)


appliance_catalog/views.py
```diff
@@ -141,8 +143,24 @@ def get_appliance_template(request, pk):
     logger.info('Getting and displaying YAML template for appliance')
 
     appliance = Appliance.objects.filter(pk=pk).get()
-    return HttpResponse(appliance.template, content_type="text/yaml")
 
+    # send message to GA
+    if settings.GOOGLE_ANALYTICS_PROPERTY_ID:
+        if request.user and request.user.username:
```
diurnalist: Do we have to attach the user's username to this? I'm not sure it's necessary, we can probably just use the random ID and not have to worry about storing things tied to user PII (personally identifiable information).

appliance_catalog/views.py
```diff
@@ -25,6 +25,8 @@
 from keystoneclient.v3 import client
 from glanceclient import Client
 from smtplib import SMTPException
+from six.moves import http_client, urllib
```
diurnalist: Why this and not `requests`? Is there a standard one we use? @xandroc

appliance_catalog/views.py
```diff
@@ -141,8 +143,24 @@ def get_appliance_template(request, pk):
...
+        params = urllib.parse.urlencode({'v': 1,
+                                         'tid': settings.GOOGLE_ANALYTICS_PROPERTY_ID,
+                                         'cid': cid,
+                                         't': 'event',
+                                         'ec': 'heat_template',
+                                         'ea': 'download',
+                                         'el': '{}-{}'.format(pk, appliance.name),
+                                         'ev': 0})
+        conn = http_client.HTTPConnection('www.google-analytics.com')
+        conn.request('POST', '/collect', params)
```
diurnalist: I'd feel more comfortable with this inside of a try/catch... otherwise this could break the appliance catalog.

zhenz:
@diurnalist GOOGLE_ANALYTICS_PROPERTY_ID is set originally for other GA trackings, and it 's only set for prod. 
https://github.com/ChameleonCloud/portal/blob/master/chameleon_env.sample#L34
https://github.com/ChameleonCloud/portal/blob/master/chameleon/settings.py#L499
https://github.com/ChameleonCloud/portal/blob/master/chameleon/templates/includes/ga.html#L1

appliance_catalog/views.py
zhenz: For users, I thought GA would have some interesting statistics, such as downloads by users. But as far as I know, there aren't any statistics on events related to users. I can just use a random id as client id.

appliance_catalog/views.py
zhenz: "six" is used somewhere else and already installed, but "requests" is not.

appliance_catalog/views.py
zhenz: will add try/catch

PR #73, opened 2019-09-27
Add new appliances endpoint returning dict with appliance uuid as key

xandroc:
Portal side of the fix for slow loading images list in Horizon

### DIFF
--- a/appliance_catalog/urls.py
+++ b/appliance_catalog/urls.py
@@ -12,6 +12,7 @@ urlpatterns = [
     url(r'^create/image/$', views.app_create_image, name='app_create_image'),
     url(r'^template/(?P<resource>.+?)\.html/$', views.app_template, name='app_template'),
     url(r'^api/appliances/$', views.get_appliances, name='get_appliances'),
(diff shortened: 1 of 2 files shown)


xandroc:
> Do you have numbers for how much this affects the performance? My intuition is that the main hit to performance is making any sort of request, and especially given that this code still has to fetch all appliances, I'm surprised that simply returning it as a dict has any impact; Python is very fast at looping over things, especially when the input size is so small (~20 objects or so.)
> 
> Approving anyways, but, I think some proof is needed for me to understand the impact here. If you have it, please include it in the PR :)

I agree, simply iterating over this short a list shouldn't cause much of a slow-down, the problem is that each request for images from Horizon was iterating over the 33 appliances for every image we have in that particular region--311 for chi@tacc, and the loop was happening 3 times for each image (once each for `get_catalog_id()` `get_is_project_supported()` and `get_is_published_in_app_catalog()`).

With the updated solution we loop over the 30-odd appliances once and generate a dict. Once the dict is generated and cached, getting appliance info by UUID is a much more efficient O(1)

diurnalist:
This was resolved in https://github.com/ChameleonCloud/horizon/pull/9

PR #75, opened 2019-10-09
permission on large uploaded files

zhenz:


zhenz:
@diurnalist I got 403 error when uploading slides to the "Talks" page. I found a fix here https://github.com/divio/django-filer/issues/1031 and tested on dev.

PR #79, opened 2019-11-18
Versioned sharing portal artifacts

diurnalist:
This adds a set of features to the sharing portal around artifact versions. Each artifact on the portal can have an arbitrary number of versions. Each version however must correspond exactly to a version already published on Zenodo, the source of record. To accomplish this, there is a new model introduced to hold the version details, and it is attached to an artifact. An artifact can have multiple of these. Each version has a DOI and the date it was created.

The versions are synced from Zenodo, either automatically (via a request to a special URL), or manually by clicking a 'Sync versions' link (which also just directs the user to this URL.) This allows us to integrate this nicely with the Jupyter environment, whereupon after a new version of a deposition is created on Zenodo, the app can redirect to this portal URL, which will then trigger the sync/update automatically for the user.

There are a lot of styling adjustments in here, and I also fixed up tests. Some functions on the artifact model were converted into properties or moved elsewhere. In particular, a Zenodo client was added to handle some of the complexities of calling the Zenodo API e.g. for fetching all versions of an artifact.


diurnalist:
@xandroc there is a lot of code in here, but almost all of it is isolated to the sharing_portal Django app. However, in 9b16883f8a4df48b255ab1d2dcc95b107aaf356b, I did remove a plugin because it was deprecated and was causing problems in the build due to one of its loose dependencies being upgraded to Python 3.

requirements.txt
```diff
@@ -12,7 +12,6 @@ django-termsandconditions
 
 # django-cms
 django-cms
-cmsplugin-filer
```
xandroc: Do you know if this will affect any existing content created with the plugin? The description of the plugin is a bit vague, but at the same time sounds like it replaced some built in functionality. `A set of cms plugins that replace the plugins shipped with django-cms with versions that use file fields from django-filer.`

diurnalist: djangocms-blog has this piece of information:

> Since cmsplugin-filer 1.1 has been moved to django-filer (since version 1.2).
https://djangocms-blog.readthedocs.io/en/latest/features/cmsplugin_filer.html#upgrading-django-cms-blog

And: https://djangocms-blog.readthedocs.io/en/latest/features/cmsplugin_filer.html#upgrading-djangocms-blog-from-0-9-x-to-1-0

So, I think this functionality is now built-in to django-filer and djangocms-picture/djangocms-file

xandroc: Perfect, this looks good then, thanks!

PR #81, opened 2020-02-06
Include interface/region in session adapter

diurnalist:
When making calls to OpenStack via its clients, it's possible to pass in
a Session object, which contains the auth information and other params.
However, you can also pass in an Adapter, which wraps a session but
pre-fills a few things that would normally be specified per-call or
per-client-construction, such as which interface/region to use (which
endpoint).

This updates all of the keystone clients to use Adapters with
interface=public and region pre-filled (where applicable.)


diurnalist:
I'm going to merge this as Portal keeps going down and we suspect this is related.

PR #83, opened 2020-04-13
Add outage email reminders

jakecoll:
Gives admins option to send outage email reminders by specified number of
seconds before outage (current set to 2 hours in settings file). This
required adding celery and redis as a broker and backend.

### DIFF
--- a/chameleon_mailman/models.py
+++ b/chameleon_mailman/models.py
@@ -6,9 +6,13 @@ from django.dispatch import receiver
 from django.utils.translation import ugettext, ugettext_lazy as _
 from django.utils.html import strip_tags
 
+from .tasks import schedule_email
+
+from datetime import timedelta
 import logging
 
 logger = logging.getLogger(__name__)
+
 
 class MailmanSubscription(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, related_name='subscriptions')
@@ -19,23 +23,32 @@ class MailmanSubscription(models.Model):
             default=True,
             help_text=_('Mailing list for discussion among Chameleon Users'))
 
+
 @receiver(post_save, sender='user_news.Outage')
 def send_outage_notification(sender, instance, using, **kwargs):
     """
     Sends email notification to the address configured in `settings.OUTAGE_NOTIFICATION_EMAIL`.
     Will send "new" notifications when the object is new.
     """
+    subject = instance.title
+    if instance.resolved:
+        subject = 'RESOLVED: %s' % subject
+
+    body = "<b>Outage Start:</b> " + instance.start_date.strftime('%Y-%m-%d %H:%M') + "<br /><br />"
+    body += "<b>Outage End:</b> " + instance.end_date.strftime('%Y-%m-%d %H:%M') + "<br /><br />"
+    body += instance.body
+    sender = settings.DEFAULT_FROM_EMAIL
+    recipients = settings.OUTAGE_NOTIFICATION_EMAIL.split(',')
+
     if instance.send_email_notification:
         logger.debug('Sending outage notification for id=%s: %s' % (instance.id, instance.title))
 
-        subject = instance.title
-        if instance.resolved:
-            subject = 'RESOLVED: %s' % subject
-
-        body = "<b>Outage Start:</b> " + instance.start_date.strftime('%Y-%m-%d %H:%M') + "<br /><br />"
-        body += "<b>Outage End:</b> " + instance.end_date.strftime('%Y-%m-%d %H:%M') + "<br /><br />"
-        body += instance.body
-        sender = settings.DEFAULT_FROM_EMAIL
-        recipients = settings.OUTAGE_NOTIFICATION_EMAIL.split(',')
-
         send_mail(subject, strip_tags(body), sender, recipients, html_message=body)
+
+    if instance.send_email_reminder:
+        eta = instance.start_date - timedelta(
+            seconds=settings.OUTAGE_EMAIL_REMINDER_TIMEDELTA)
+        subject = 'Outage Reminder: {}'.format(subject)
+        schedule_email.apply_async(
+            (subject, strip_tags(body), sender, recipients),
+            kwargs=dict(html_message=body), eta=eta)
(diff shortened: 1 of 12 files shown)


chameleon_mailman/models.py
```diff
@@ -6,10 +6,17 @@
...
 
+from .tasks import schedule_email
+
+from datetime import timedelta
 import logging
 
 logger = logging.getLogger(__name__)
 
+
+SECONDS_BEFORE_REMINDER_EMAIL = (60 * 60 * 2)
```
diurnalist, line 17: Should this read from the `OUTAGE_EMAIL_REMINDER_TIMEDELTA` defined in settings?

chameleon_mailman/tasks.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import absolute_import, unicode_literals
+
+from celery import shared_task
+
+from django.core.mail import send_mail
+
+
+@shared_task
+def schedule_email(*args, **kwargs):
```
diurnalist, line 9: Nice, that is very simple 👍 - is there any befit in returning the result?

requirements.txt
```diff
@@ -45,3 +45,7 @@ python-glanceclient
 python-keystoneclient
 python-novaclient
 python-blazarclient
+
+celery
```
diurnalist, line 49: Is this the celery internal library, or just a client to celery? I think we would only need the client here.

requirements.txt
```diff
@@ -45,3 +45,7 @@ python-glanceclient
...
+redis
```
diurnalist, line 50: Why do we need this?

user_news/forms.py
```diff
@@ -21,3 +22,8 @@ def clean_send_email_notification(self):
         send = self.cleaned_data.get('send_email_notification')
         self.instance.send_email_notification = send
         return send
+
+    def clean_send_email_reminder(self):
+        send = self.cleaned_data.get('send_email_reminder')
```
diurnalist, line 27: Just curious, do you know why we do this/what it does?

user_news/models.py
```diff
@@ -74,6 +74,7 @@ class Outage(News):
     end_date = models.DateTimeField('expected end of outage')
     resolved = models.BooleanField('resolved', default=False)
     send_email_notification = False
+    send_email_reminder = False
```
diurnalist, line 77: We will also need a DB migration for this. I don't remember the exact invocation to do this, but I think Django makes it pretty easy to auto-generate them!

chameleon_mailman/models.py
jakecoll: Moved this to settings in clean up and forgot to remove old code.

chameleon_mailman/tasks.py
jakecoll: No, on the contrary, returning was breaking my code when testing. It seems apply_async was trying to operate on an int object if return was added. It is also possible I was using apply_async incorrectly at that time..

requirements.txt
jakecoll: The space? There seems to be some kind of organizing structure in the requirements text. I moved celery, redis, and kombu up with the list of non-openstack and non-django related requirements.

user_news/forms.py
jakecoll: This is a good question and it took me awhile to figure it out. From what I can tell, this allows us to override needing to store the send_email_notification and send_email_reminder in the database (hence no migration needed). We have the email notification set to check and send an email whenever the content of an Outage form are set to be saved to the database. The check boxes are added to the form, but they are not model fields and therefore will not be stored in the db. Instead, we override django's cleaned data functionality to add the contents of the form to the model instance so that we can determine if we send an email without then having to store it in the database.

user_news/models.py
jakecoll: Based on my previous comment above, it's unclear if we need these here. We do not store send_email_notification or send_email_reminder boolean values in the database so no migration is needed.

user_news/forms.py
diurnalist: Ah, nice explanation, thanks! OK, this makes sense. I guess there is not much value in saving the flag in the DB. I initially thought that maybe we would store something like, the date the outage reminder was sent, and if it's NULL then it knows to send the reminder when periodically checking if any reminders should be sent. But, since we are scheduling the emails to occur at a precise time, there isn't much need to store it.

This actually just occurred to me while writing this... because we are scheduling the emails at a future dates, I think we might have an issue where the emails get sent at the wrong time if the outage date is ever changed. Is there a way to clear out a scheduled task? Maybe we can do something where, if the outage is ever updated, the scheduled task is cleared and replaced with a new scheduled task?

diurnalist: I'm wondering if we actually need to store the celery task ID in order to do this. Maybe we do need to store a field in the DB, like `notification_task_id`. I think the return of `apply_async` is something you can get a task ID from. Then, when updating an outage, you can revoke the task given its ID, create a new task, and store the new task ID. An optimization would be to not do any of this if the outage start date did not change.

requirements.txt
jakecoll: I assume you are referring to the kombu? It's a transport library for celery. I believe we need it. I spent some time trying to figure out a compatible version for something that was breaking while testing.

user_news/forms.py
jakecoll: Looking into this a little deeper I think it can be done possibly one of three ways. The last two ways depend on if the signal is sent before or after the post is saved.

1) We assign the task_id and make it the same at the Outage model id. This way we can call the model id and tell celery to purge any tasks with that id before assigning a new task with that id. This would allow us to avoid migration.
2) If signal sent after post is saved. We update a new field that we exclude from the form that holds the email_reminder_task_id with the celery task id.
3) If signal sent before post is saved. We call the process in the send_email_reminder task in the model on save and apply the task id to the field.

jakecoll: I took a bit of a hacky approach. I set the task id of the schedule_email task and attach a version number. This allows us to revoke the old task and create a new task with an incremented version number.

PR #85, opened 2020-05-05
manage allocations using portal db

zhenz:


allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
+import requests
+import json
+from django.conf import settings
+import logging
+from keystoneauth1.identity import v3
+from keystoneauth1 import session
+
+logger = logging.getLogger('default')
```
diurnalist, line 8:
```suggestion
logger = logging.getLogger('allocations')
```

allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
...
+
+class BALANCE_SERVICE_API:
```
diurnalist, line 10: It's not typical to name Pyton classes LIKE_THIS - that convention is reserved for constants. How about `BalanceServiceClient`?

allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
...
+        auth = v3.Password(auth_url=settings.OPENSTACK_KEYSTONE_URL,
+                           username=settings.OPENSTACK_SERVICE_USERNAME,
+                           password=settings.OPENSTACK_SERVICE_PASSWORD,
+                           project_id=settings.OPENSTACK_SERVICE_PROJECT_ID,
+                           user_domain_name='default',
+                           user_domain_id='default')
+        sess = session.Session(auth=auth)
+        auth_token = sess.get_token()
+        
+        headers = {'Accept': 'appliaction/json',
```
diurnalist, line 23:
```suggestion
        headers = {'Accept': 'application/json',
```

Sometimes it's nice to take a string that can be misspelled and pull it out to its own constant:

```python
CONTENT_TYPE_JSON = 'application/json'
```

It's possible such a constant already exists in one of the libraries we use.

allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
...
+    def call( self, project_code ):
+        # get token
+        auth = v3.Password(auth_url=settings.OPENSTACK_KEYSTONE_URL,
+                           username=settings.OPENSTACK_SERVICE_USERNAME,
+                           password=settings.OPENSTACK_SERVICE_PASSWORD,
+                           project_id=settings.OPENSTACK_SERVICE_PROJECT_ID,
+                           user_domain_name='default',
+                           user_domain_id='default')
+        sess = session.Session(auth=auth)
+        auth_token = sess.get_token()
```
diurnalist, line 21: I think there is even a `get_headers` or something, which might be the more future-proof solution. It would return a headers dict, which you can then add your content-type stuff to.

allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
...
+        headers = {'Accept': 'appliaction/json',
+                   'X-Auth-Token': auth_token,
+                   'Content-Type': 'appliaction/json'}
+        
+        url = self.make_url( project_code )
+        logger.info('Requesting %s from allocations balance service ...', url)
+        resp = requests.get(url, headers=headers)
+        logger.info('Response received. Parsing to json ...')
+        
+        data = json.loads(resp.text)
```
diurnalist, line 32: You can just do `resp.json()` I think

allocations/allocations_api.py
```diff
@@ -0,0 +1,36 @@
...
+        return data
+
+    def make_url( self, project_code ):
+        return '{0}/{1}'.format( settings.ALLOCATIONS_BALANCE_SERVICE_ROOT_URL, project_code)
```
diurnalist, line 36: Newline at end of file

allocations/models.py
```diff
@@ -0,0 +1,28 @@
+from django.db import models
+import json
+import logging
+from django.conf import settings
+
+logger = logging.getLogger('allocations')
+
+class Allocation(models.Model):
+    STATUS = (
+        ('Inactive', 'Expired or overused'),
```
diurnalist, line 10: what about making these uncapitalized? It makes it easier to normalize them when storing:

```python
status = req.params['status'].lower()
```

allocations/models.py
```diff
@@ -0,0 +1,28 @@
...
+        ('Active', 'Active'),
+        ('Pending', 'Waiting for review decision'),
+        ('Rejected', 'Rejected'),
+        ('Approved', 'Approved but not active')
+    )
+    project_charge_code = models.CharField(max_length=50,blank=False)
+    status = models.CharField(max_length=50, blank=False,choices=STATUS)
+    justification = models.CharField(max_length=500,null=True)
+    requestor = models.ForeignKey(settings.AUTH_USER_MODEL,related_name='allocation_requestor', null=True)
```
diurnalist, line 19: It's really nice to be able to do this. I suppose it means all reviewers have to have Chameleon accounts, but that is a very reasonable requirement.

allocations/allocations_api.py
diurnalist, line 36: What if ALLOCATIONS_BALANCE_SERVICE_ROOT_URL is `None` or empty string? Or otherwise malformed? Where does it make sense to raise that error?

chameleon/templates/dashboard.html
```diff
@@ -70,18 +70,18 @@ <h4 class="panel-title pull-left">Active Projects</h4>
                       </td>
                       <td>
                         {% for a in project.allocations %}
-                          {% if a.resourceId == alloc.resourceId %}
```
diurnalist, line 73: This is going to break compatibility with the current UI, which works against TAS. We need to either make this UI work for both somehow (maybe shimming the TAS model with attributes like `date_reviewed` which map to their internal properties (`dateReviewed`)? Happy to discuss in more detail.

projects/templates/projects/allocation.html
```diff
@@ -1,6 +1,5 @@
 <div class="allocation allocation-{{ alloc.status|lower }}" data-allocation-id="{{ alloc.id }}" data-allocation-status="{{ alloc.status|lower }}">
   <h4>
-    {{ alloc.resource }}
```
diurnalist, line 3: ?

projects/views.py
```diff
@@ -165,40 +166,46 @@ def view_project(request, project_id):
...
     project.inactive_allocations = []
 
-    for a in project.allocations:
-        if a.status == 'Active' and a.resource == 'Chameleon':
+    project_allocations = Allocation.objects.filter(project_charge_code=project.chargeCode)
+
+    for a in project_allocations:
+        if a.status == 'Active':
             project.active_allocations.append(a)
             project.has_active_allocations = True
```
diurnalist, line 174: General note on this: I think it would really clean up this code to make some of these "derived" properties actually derived from other state. For example, on the Project:

```python
@property
def has_active_allocations(self):
    return len(self.active_allocations) > 0

@property
def has_inactive_allocations(self):
    return len(project.inactive_allocations) > 0
```

Similarly, on your Allocation model:

```python
@property
def days_left(self):
    now = datetime.utcnow().replace(tzinfo=None)
    return (self.expiration_date.replace(tzinfo=None) - now).days

@property
def percent_su_used(self):
    if self.su_allocated and self.su_used:
        return self.su_used / self.su_allocated * 100
    else:
        return 0.0
```

This way, you only have to manage the "core" values of state, and the rest can be computed on-the-fly, as it "derives" from the core state naturally via some transformation, which is fixed.

allocations/models.py
```diff
@@ -0,0 +1,28 @@
+from django.db import models
+import json
+import logging
+from django.conf import settings
+
+logger = logging.getLogger('allocations')
+
+class Allocation(models.Model):
```
xandroc, line 8: can you check-in the migration generated by this model, should be something like `allocations/migrations/0001_initial.py`, migrations can get tangled up otherwise, checking them in helps to make sure everyone's on the same page

zhenz:
@diurnalist @xandroc I created a mapping that maps from portal allocation model to tas allocation model. And to use data from where is determined by "ALLOCATIONS_FROM_PORTAL" env variable. Some changes are made in pytas repo. I created a PR for pytas changes as well.

allocations/models.py
```diff
@@ -0,0 +1,41 @@
+from django.db import models
+import json
+import logging
+from django.conf import settings
+from datetime import datetime
+
+logger = logging.getLogger('allocations')
+
+class Allocation(models.Model):
```
xandroc, line 9: When running this locally and trying to view a project `/user/projects/25142/`, I get the error below,
```
Exception Value: | 'Allocation' object has no attribute 'days_left'
-- | --
/project/projects/views.py in view_project, line 178
```

zhenz: @xandroc You need to use "allocations" branch of pytas. I had a PR for the changes: https://github.com/ChameleonCloud/pytas/pull/1

allocations/migrations/0001_initial.py
```diff
@@ -0,0 +1,38 @@
...
+                ('status', models.CharField(choices=[(b'inactive', b'expired or overused'), (b'active', b'active'), (b'pending', b'waiting for review decision'), (b'rejected', b'rejected'), (b'approved', b'approved but not active')], max_length=50)),
+                ('justification', models.CharField(max_length=500, null=True)),
+                ('date_requested', models.DateTimeField(null=True)),
+                ('decision_summary', models.CharField(max_length=500, null=True)),
+                ('date_reviewed', models.DateTimeField(null=True)),
+                ('expiration_date', models.DateTimeField(null=True)),
+                ('su_requested', models.FloatField()),
+                ('start_date', models.DateTimeField()),
+                ('su_allocated', models.FloatField()),
+                ('su_used', models.FloatField()),
```
diurnalist, line 33: I'm not sure we should be storing this here - this is what the balance service is for, no? The `su_requested` and `su_allocated` make sense as they can be some sort of indicator of what the Chameleon approver did with the request, but I think `su_used` should be the balance service's responsibility to track.

allocations/allocations_api.py
```diff
@@ -0,0 +1,31 @@
...
+import traceback
+from django.conf import settings
+import logging
+from keystoneauth1.identity import v3
+from keystoneauth1 import session
+from project.view import get_ks_auth_and_session
+
+logger = logging.getLogger('allocations')
+
+class BalanceServiceClient:
```
diurnalist, line 11: Where is this used? I didn't see any reference to it.

I'd also recommend doing a bit of checking in an `__init__()` function to see if the settings are set up. It should raise an error if there is no `ALLOCATIONS_BALANCE_SERVICE_ROOT_URL`, as that looks to be required.

allocations/allocations_api.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def call( self, project_code ):
+        auth, sess = get_ks_auth_and_session()
+        headers = auth.get_headers(sess)
+        headers['Accept'] = 'application/json'
+        
+        url = self.make_url( project_code )
+        logger.info('Requesting %s from allocations balance service ...', url)
```
diurnalist, line 19: I know I am very guilty of writing log messages like this, saying "I'm _about_ to do something", but I don't think it's the right way to log for this type of thing. Logging based on intent is a bit dangerous because if you're just looking at the logs, you have no idea if the code _actually_ did what was intended. Instead, I recommend doing this style:

```python
logger.info('Successfully retrieved balance for project %s', project_code)
logger.exception('Failed to retrieve balance for project %s', project_code)
```

If you want to include debug information, you can do it on another verbosity level:

```python
logger.debug('Response from %s: %s', url, data)
```

This keeps INFO logs clean (which they should be) - INFO should be telling you important operational information about what the system is doing. DEBUG should be for when you need to figure out why something is _not_ working properly. And of course ERROR is for when things fail.

allocations/allocations_api.py
```diff
@@ -0,0 +1,31 @@
...
+        data = None
+        try:
+            resp = requests.get(url, headers=headers)
+            logger.info('Response received. Parsing to json ...')
+            data = resp.json()
+        except Exception:
+            logger.error(traceback.format_exc())
```
diurnalist, line 26:
```suggestion
            logger.exception('Failed to retrieve balance for project %s', project_code)
```

allocations/allocations_api.py
```diff
@@ -0,0 +1,31 @@
...
+
+        return data
+
+    def make_url( self, project_code ):
+        return '{0}/{1}'.format( settings.ALLOCATIONS_BALANCE_SERVICE_ROOT_URL, project_code)
```
diurnalist, line 31: nit: remove space

allocations/models.py
```diff
@@ -0,0 +1,41 @@
...
+    STATUS = (
+        ('inactive', 'expired or overused'),
```
diurnalist, line 11: Does the second item here map to a UI element somehow? I presume it's not stored in the DB?

projects/views.py
```diff
@@ -195,10 +175,9 @@ def view_project(request, project_id):
             if isinstance(a.end, basestring):
                 a.end = datetime.strptime(a.end, '%Y-%m-%dT%H:%M:%SZ')
 
-            days_left = (a.end - datetime.today()).days
-            if days_left >= 0 and days_left <= 90:
+            if a.days_left >= 0 and a.days_left <= 90:
                 a.up_for_renewal = True
```
diurnalist, line 179: This also looks like a derived property :)

```python
RENEWAL_START_WINDOW = 90

@property
def up_for_renewal(self):
    days_left = self.days_left
    return days_left >= 0 and days_left <= RENEWAL_START_WINDOW
```

And in that case, no need to deal with this `renewal_days`, just use `days_left` - I don't know if another property helps, given it always has the same value.

chameleon/views.py
```diff
@@ -231,7 +232,8 @@ def dashboard(request):
             proj.__dict__['nickname'] = extras.nickname
         except ProjectExtras.DoesNotExist:
             project_nickname = None
-
+        
+        proj = project_allocations_format_mapping(proj)
```
diurnalist, line 236: How could we make the decision of "fetch allocations from portal... or not" be based on the incoming request? I'd like to make it so we can roll this out in "dark launch", meaning most users don't interact with it at all.

I suggest you make this mapping thing a class, which is instantialized with a request, e.g.

```python
class ProjectAllocationMapper:
    def __init__(self, request):
        self.allocations_from_db = self._wants_db_allocations(request)
    
    def map(self, project):
        if self.allocations_from_db:
            # Do your fetching and mapping of allocations
            some_logic
        else:
            return project
```

Then in your controller (`dashboard`), you can do like:

```python
mapper = ProjectAllocationsMapper(request)

proj = mapper.map(proj)
```

Then we can vary the logic of `_wants_db_allocations` to look at a cookie, or perhaps the logged in user ID, or something else entirely (like hash the user ID and then do some bucketing such that we roll out to 50% of users, then 100%)

projects/views.py
```diff
@@ -159,30 +160,9 @@ def view_project(request, project_id):
     if not project_member_or_admin_or_superuser(request.user, project, users):
         raise PermissionDenied
 
-    project.active_allocations = []
-    project.approved_allocations = []
-    project.pending_allocations = []
-    project.rejected_allocations = []
-    project.inactive_allocations = []
+    project = project_allocations_format_mapping(project)
```
diurnalist, line 163: Same comment as above applies here.

allocations/models.py
zhenz: no. it's not mapped to UI element and not stored in the db. but django `choices` requires two items.

allocations/allocation_mapper.py
```diff
@@ -0,0 +1,55 @@
+from pytas.models import Allocation as tas_alloc
+from allocations.models import Allocation as portal_alloc
+from allocations.allocations_api import BalanceServiceClient
+
+class ProjectAllocationMapper:
+    def __init__(self, request):
+        self.allocations_from_db = self._wants_db_allocations(request)
+        
+    def _wants_db_allocations(self, request):
+        if request.user.is_superuser:
```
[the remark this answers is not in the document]
zhenz: You can go to the portal admin page and select "Users". Users with "staff status" are superusers. We have 25 superusers now.

PR #86, opened 2020-05-14
Outage reminder fix

jakecoll:


chameleon_mailman/tasks.py
```diff
@@ -1,10 +1,38 @@
...
 from django.core.mail import send_mail
+from django.utils.html import strip_tags
+from user_news.models import Outage
+
+logger = logging.getLogger(__name__)
+
+
+@task
+def send_outage_reminders(crontab_frequency, send_outage_reminder_before):
+    now = pytz.utc.localize(datetime.utcnow())
```
diurnalist, line 19: Why is this localizing to the system time? Is the system time not UTC? Is the database TZ not UTC?

I always have to look up how to do this. Apparently "the way" from what I have read is `datetime.now(pytz.utc)`

chameleon/settings.py
```diff
@@ -639,3 +642,12 @@
 CELERY_RESULT_SERIALIZER = 'json'
 CELERY_TASK_SERIALIZER = 'json'
 OUTAGE_EMAIL_REMINDER_TIMEDELTA = (60 * 60 * 2)
+OUTAGE_REMINDER_FREQUENCY = 30
```
diurnalist, line 645: I'd prefer we keep all of these in the same unit - can you make this seconds? You can keep the cron schedule in minutes, just divide by 60 (and cast to `int`!). It makes it easier to compare the deltas w/ the frequency when examining the config. No surprises.

chameleon_mailman/tasks.py
```diff
@@ -1,10 +1,38 @@
...
+    upcoming_outages = Outage.objects.filter(start_date__gte=now)
+
+    send_window_max = now + timedelta(seconds=send_outage_reminder_before)
+    send_window_min = send_window_max - timedelta(
+        seconds=(crontab_frequency * 60))
+
+    for outage in upcoming_outages.all():
+        if send_window_min < outage.start_date < send_window_max:
```
diurnalist, line 27: The frequency is 30 minutes, but the total window is 2 hours. Unless we somehow store that we've already sent mail, this will send 4 mails per outage, no? That's why I was advocating for adding a model field to store something like `notification_sent_at`, storing a date of when the notification was sent (better for auditing than a simple boolean `notification_sent`, which only tells you if it was sent or not).

diurnalist: You could test this by setting interval very low, like 1 minute, and the "remind before" window to 5 minutes. I'd expect 5 mails to be generated from the test environment.

docker-compose.yml
```diff
@@ -1,32 +1,34 @@
...
+      - 8000:8000
+  redis:
+    image: "redis:alpine"
+    expose:
+      - '6379'
+    ports:
+      - '6379:6379'
+  celery:
+    build: .
+    command: celery -A chameleon worker -l info --concurrency=1 -B
```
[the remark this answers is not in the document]
jakecoll: The -B (--beat) has the worker run a beat scheduler.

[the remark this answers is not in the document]
diurnalist: If I look up the [docs](https://docs.celeryproject.org/en/stable/reference/celery.bin.worker.html#cmdoption-celery-worker-b), there is a big note on this flag:

> -B is meant to be used for development purposes. For production environment, you need to start celery beat separately.

How difficult is it to start the beat separately? "for development purposes" can mean it's sort of OK if you're not expecting high load (which we're not), but it can also mean "we don't really test or support this in the same way, and it has these failure modes", which wouldn't be OK.

chameleon_mailman/tasks.py
```diff
@@ -1,10 +1,44 @@
...
+            sender = settings.DEFAULT_FROM_EMAIL
+            recipients = settings.OUTAGE_NOTIFICATION_EMAIL.split(',')
 
+            mail_sent = send_mail(
+                subject, strip_tags(body), sender, recipients)
 
-@shared_task
-def schedule_email(*args, **kwargs):
-    send_mail(*args, **kwargs)
+            if mail_sent:
```
diurnalist, line 42: Perfect 👍 - you could even change the type to a datetime if you want to store when it was sent

docker-compose.dev.yml
```diff
@@ -2,7 +2,7 @@ version: '3.3'
 
 services:
   portal:
-    image: docker.chameleoncloud.org/portal:latest
+    build: .
```
diurnalist, line 5: Good improvement

docker-compose.yml
```diff
@@ -1,32 +1,34 @@
...
-celery:
-  build: .
-  command: celery -A chameleon worker -l info
-  env_file:
-    - .chameleon_env
-  volumes:
-    - .:/project
-  depends_on:
-    - redis
+version: '3.3'
```
diurnalist, line 1: Weirdly, I don't think this docker-compose file is actually used in prod (is it?) I imagine the celery stuff had to be copied over, no?

user_news/forms.py
```diff
@@ -12,18 +12,13 @@ class Meta:
 
 class OutageForm(ModelForm):
     send_email_notification = BooleanField(required=False)
-    send_email_reminder = BooleanField(required=False)
 
     class Meta:
         model = Outage
+        exclude = ['reminder_sent']
```
diurnalist, line 18: Nice, didn't know about this thing. https://docs.djangoproject.com/en/1.11/topics/forms/modelforms/#selecting-the-fields-to-use

PR #87, opened 2020-05-26
Recharge/create allocation

zhenz:


allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
-    def map(self, project):
+    def map(self, project, get_redis_data = True):
         if self.allocations_from_db:
+            is_dict_type = isinstance(project, dict)
             balance_service = BalanceServiceClient()
             reformated_allocations = []
-            for alloc in portal_alloc.objects.filter(project_charge_code=project.chargeCode):
-                if alloc.status == 'active':
-                    active_allocation = balance_service.call(project.chargeCode)
+            project_charge_code = project['chargeCode'] if is_dict_type else project.chargeCode
```
diurnalist, line 42: Another way of doing this:

```python
def get_attr(obj, key):
    """Attempt to resolve the key either as an attribute or a dict key"""
    if isinstance(obj, dict):
        return obj.get(key)
    else:
        return getattr(obj, key, None)

# Later...
project_charge_code = get_attr(project, 'chargeCode')
```

You can then use that instead of having the `is_dict_type` stuff peppered around here.

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+            for alloc in portal_alloc.objects.filter(project_charge_code=project_charge_code):
+                if get_redis_data and alloc.status == 'active':
+                    active_allocation = balance_service.call(project_charge_code)
```
diurnalist, line 45: We are calling the balance service again here unnecessarily; I didn't catch this on the first go. Suggestion to tidy this up, and also to add logging around weird responses from the balance service:

```suggestion
                    balance = balance_service.call(project_charge_code)
                    su_used = balance.get('used')
                    if su_used is not None:
                        alloc.su_used = float(su_used)
                    else:
                        # This is an error that we should log; the balance service may be acting up
```

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+            else:
+                project.allocations = reformated_allocations
         return project
+    
+    def save_allocation(self, alloc, project_charge_code, tas):
+        if self.allocations_from_db:
+            reformated_alloc = self.tas_to_portal_obj(alloc, project_charge_code)
+            reformated_alloc.save()
+            # notify Chameleon operators
+            subject = 'Pending allocation request for project {}'.format(project_charge_code)
```
diurnalist, line 63: Suggest moving the email sending to its own private function:

```python
def _send_allocation_request_notification(self, charge_code):
    subject = 'Pending allocation request for project {}'.format(charge_code)
    body = '...'
    return send_mail(subject, strip_tags(body), ...)
```

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+    def get_username(self, request, tas):
+        if self.allocations_from_db:
+            return request.user.username
+        else:
+            return tas.get_user(username=request.user)['username']
+
+    def allocation_approval(self, data, tas):
+        if self.allocations_from_db:
+            # update allocation model
+            alloc = portal_alloc.objects.filter(id=data['id'])[0]
```
diurnalist, line 87: When retrieving by primary key (`id` in this case), there is a better way: https://docs.djangoproject.com/en/1.11/topics/db/queries/#retrieving-a-single-object-with-get

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+            data['status'] = data['status'].lower()
+            for item in ['reviewerId', 'dateReviewed', 'start', 'end', 'status', 'decisionSummary', 'computeAllocated']:
+                setattr(alloc, TAS_TO_PORTAL_MAP[item], data[item])
+            alloc.save()
+            logger.info('Allocation model updated: data=%s', alloc.__dict__)
+            # send email to PI
+            user_model = get_user_model()
+            user = user_model.objects.filter(id=data['requestorId'])[0]
+            subject = 'Decision of your allocation request for project {}'.format(data['project'])
```
diurnalist, line 96: Would also clean up this function if the email sending was a private helper function

allocations/views.py
```diff
@@ -49,6 +50,8 @@ def view( request ):
         tas = TASClient()
         resp = tas.projects_for_group('Chameleon')
         for p in resp:
+            mapper = ProjectAllocationMapper(request)
+            p = mapper.map(p, get_redis_data = False)
```
diurnalist, line 54: Nitpick: after seeing this in a few places, I'm thinking that `fetch_balance` is probably a better name for this argument. The fact that it's Redis shouldn't really matter (and in fact, we are getting it from a balance service anyways, not Redis directly).

allocations/views.py
```diff
@@ -90,7 +93,7 @@ def user_projects( request, username ):
                 temp = {}
                 # run through and make sure all the allocations are associated with one project
                 for p in userProjects:
-                     if p['source'] == 'Chameleon':
+                    if p['source'] == 'Chameleon':
```
diurnalist, line 96: How long has this been broken?

allocations/views.py
```diff
@@ -121,10 +124,10 @@ def approval( request ):
     status = ''
     if request.POST:
         tas = TASClient()
-        userData = tas.get_user( username=request.user )
+        mapper = ProjectAllocationMapper(request)
         data = json.loads(request.body)
-        data['reviewer'] = userData['username']
-        data['reviewerId'] = userData['id']
+        data['reviewer'] = mapper.get_username(request, tas)
```
diurnalist, line 129: Very nice 👍

util/scripts/sync_allocation_from_tas.conf
```diff
@@ -0,0 +1,9 @@
+[tas]
+url=
+client_key=
+client_secret=
+
+[portal]
+host=
+username=
+password=
```
diurnalist, line 9: Please add a newline here - it is possible to configure IDEs to do this automatically for you

util/scripts/sync_allocation_from_tas.py
```diff
@@ -0,0 +1,158 @@
...
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+TAS_DATE_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+
+PORTAL_ALLOCATION_TABLE_NAME = 'allocations_allocation'
+PORTAL_AUTH_USER_TABLE_NAME = 'auth_user'
+
+TAS_TO_PORTAL_MAP = {'project': 'project_charge_code',
```
diurnalist, line 17: Does it make sense to `import` this somehow? Then we don't have to worry about it being out of sync. Same applies for TAS_DATE_FORMAT

allocations/templates/allocations/approval.html
```diff
@@ -31,7 +31,7 @@ <h3 class="modal-title">Allocation Approval</h3>
             <tr ng-show="data.allocation.computeRequested && data.allocation.computeRequested > 0 && data.type == 'approve'">
                 <td><strong>Compute Allocated:</strong>&nbsp;<sup><i style="color: red;font-size:0.5em;" class="fa fa-asterisk"></i></sup></td>
                 <td>
-                    <input type="number" name="computeAllocated" ng-model="data.allocation.computeAllocated" ng-required="true" value=''>
+                    <input type="number" name="computeAllocated" ng-model="data.allocation.computeAllocated" ng-required="data.type == 'approve'" value=''>
```
diurnalist, line 34: What is this change doing? It was a bit hard to follow how this `type` comes in here

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+            for alloc in portal_alloc.objects.filter(project_charge_code=project_charge_code):
+                if get_redis_data and alloc.status == 'active':
+                    active_allocation = balance_service.call(project_charge_code)
                     if 'used' in active_allocation and active_allocation['used']:
-                        alloc.su_used = float(balance_service.call(project.chargeCode)['used'])
-                reformated_allocations.append(self.portal_to_tas_obj(alloc))
-            project.allocations = reformated_allocations
+                        alloc.su_used = float(balance_service.call(project_charge_code)['used'])
+                reformated_alloc = self.portal_to_tas_obj(alloc)
+                if is_dict_type:
```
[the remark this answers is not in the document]
zhenz: This is not for `__dict__`. There are several places that use this function, and depends on what `tasclient` method is used, it may have "project" as pytas.project object or a dict type.

allocations/allocation_mapper.py
```diff
@@ -11,45 +34,129 @@ def _wants_db_allocations(self, request):
...
+        if self.allocations_from_db:
+            return request.user.id
+        else:
+            return tas.get_user(username=request.user)['id']
+
+    def get_username(self, request, tas):
+        if self.allocations_from_db:
+            return request.user.username
+        else:
+            return tas.get_user(username=request.user)['username']
```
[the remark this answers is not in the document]
zhenz: there's no difference. will update.

util/scripts/sync_allocation_from_tas.py
zhenz: I don't think so...

allocations/templates/allocations/approval.html
zhenz: It's defined in the controller https://github.com/ChameleonCloud/portal/blob/master/allocations/scripts/controllers/controller.js#L130-L134. Type can be either "approve" or "reject". And if you click "reject", some fields will be hide. See the screenshot below.

Screen Shot 2020-06-02 at 12 37 15 PM

Before fix, the fields like "compute allocated" will be hidden, but still required, which makes the "submit" button unable to click. So whether the field is required or not depends on the "type" as well.

zhenz:
@diurnalist Please let me know if any of my replies don't make sense to you. Also, I had periodic activate/expire task in "allocations/tasks.py". It's pushed after your first review.

allocations/allocation_mapper.py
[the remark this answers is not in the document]
diurnalist: OK, feel free to resolve when the change is in :)

allocations/tasks.py
```diff
@@ -0,0 +1,92 @@
...
+    # activate allocations
+    approved_allocations = Allocation.objects.filter(status='approved', start_date__lte=now)
+    activated_alloc_count = 0
+    for alloc in approved_allocations:
+        # deactivate active allocation for the project and set status to active
+        project_active_allocations = Allocation.objects.filter(status='active', project_charge_code=alloc.project_charge_code)
+        proj_alloc = None
+        if len(project_active_allocations) > 0:
+            proj_alloc = project_active_allocations[0]
+        with transaction.atomic():
```
diurnalist, line 67: I like this approach. I think we can make it even safer, while keeping it quite simple:

```python
try:
    with transaction.atomic():
        su_used = float(balance_service.call(alloc.project_charge_code)['used'])
        if proj_alloc:
            proj_alloc.su_used = su_used
            proj_alloc.status = 'inactive'
            proj_alloc.save()
        alloc.status = 'active'
        alloc.save()
        balance_service.recharge(alloc.project_charge_code, alloc.su_allocated)
except Exception:
    log.exception('Error activating project {}'.format(alloc.project_charge_code))
```

This should accomplish the following:
- if we fail to fetch the balance, nothing happens
- if one of the DB transactions fails, the transaction rolls back
- if we fail to recharge the allocation, the transaction rolls back
- if any allocation update fails, the system still proceeds to the next one in the list for processing (it doesn't abort)

I think this should ensure that we keep the two systems in sync, if I'm reasoning about it correctly.

allocations/tasks.py
```diff
@@ -0,0 +1,92 @@
...
+                    proj_alloc.su_used = None
+                    proj_alloc.status = 'active'
+                    proj_alloc.save()
+                alloc.status = 'approved'
+                alloc.save()
+                logger.info('successfully roll back portal db change for {}'.format(alloc.project_charge_code))
+            except Exception as re:
+                logger.exception(re)
+                
+    logger.debug('need to activated {} allocations, and {} were actually activated'.format(len(approved_allocations), activated_alloc_count))
```
diurnalist, line 92: This is a long task, it may be easier to read if the 3 different "subtasks" are expressed as their own private function.

zhenz:
> See notes about transactions. I really like the usage of `atomic()` that you have in the recharge; I think we should take that approach for reset as well. The periodic task overall looks good, just left some comments. I'm not sure if some of the other feedback I left still applies, sometimes GitHub's UI is confusing and shows me files from old commits.

@diurnalist I think I fixed all of your comments :)

allocations/tasks.py
```diff
@@ -0,0 +1,83 @@
...
+    now = datetime.now(pytz.utc)
+
+    approved_allocations = Allocation.objects.filter(status='approved', start_date__lte=now)
+    activated_alloc_count = 0
+    for alloc in approved_allocations:
+        # deactivate active allocation for the project and set status to active
+        project_active_allocations = Allocation.objects.filter(status='active', project_charge_code=alloc.project_charge_code)
+        proj_alloc = None
+        if len(project_active_allocations) > 0:
+            proj_alloc = project_active_allocations[0]
```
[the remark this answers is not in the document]
zhenz: yes.

PR #88, opened 2020-05-27
Adding Keycloak Federated Auth to Portal

xandroc:
Integrating Portal and Keycloak auth using OIDC. User creation and update sets username, first name, and last name


requirements.txt
```diff
@@ -43,6 +43,7 @@ slimit
 sqlparse
 pytas
 bibtexparser
+mozilla-django-oidc
```
diurnalist, line 46: We might need an upper-constraint on this, given that we only support Django 1.11 at the moment. Is there a version of this module that drops support for Django 1.11/Python 2?

chameleon/settings.py
```diff
@@ -111,6 +112,7 @@
     ##
     # custom
     #
+    'chameleon',
```
diurnalist, line 115: Is it OK for a Django site to load itself as an app? I thought that there was some difference in purpose between the "site" directory (./chameleon) and the app directories (e.g., ./chameleon_openid)

chameleon/migrations/0001_initial.py
```diff
@@ -0,0 +1,26 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('auth', '0008_alter_user_username_max_length'),
```
diurnalist, line 15: What migration is this? It does make me think that we are going to have to loosen username length requirements when we start accepting accounts other than TACC ones.

chameleon/ChameleonOIDCAuthBackend.py
```diff
@@ -0,0 +1,37 @@
+from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+from chameleon.models import UserProperties
+import logging
+
+logger = logging.getLogger('auth')
+
+class ChameleonOIDCAB(OIDCAuthenticationBackend):
+    def create_user(self, claims):
+        if self.get_settings('OIDC_USERNAME_ALGO', None):
```
diurnalist, line 9: I don't quite get what this branch is doing - why do we check this particular setting and skip the rest of the flow here?

chameleon/settings.py
```diff
@@ -250,9 +252,19 @@
     os.path.join(BASE_DIR, 'locale'),
 )
 
+## Keycloak OIDC Authentication
+OIDC_RP_CLIENT_ID = os.environ.get('OIDC_RP_CLIENT_ID')
+OIDC_RP_CLIENT_SECRET = os.environ.get('OIDC_RP_CLIENT_SECRET')
+OIDC_RP_SIGN_ALGO = os.environ.get('OIDC_RP_SIGN_ALGO')
+OIDC_OP_JWKS_ENDPOINT = os.environ.get('OIDC_OP_JWKS_ENDPOINT')
```
diurnalist, line 259: Are all of these endpoints necessary? I'm not sure that we need the JWKS endpoint, for example.

requirements.txt
xandroc: > The changes look pretty simple to me. How can we differentiate between users who logged in via OpenID versus the ones who were pre-existing / haven't logged in with that method?
> 
> What about the changes for things like, is PI status approved, who approved it and when? Is that going to be in another PR?
> 
> What do you think about renaming `is_pi_eligible` to `is_pi`? If the role is on the user in Keycloak, that will meant they have been granted the status.

I'm still working on differentiating between the two types of login methods. I was planning on making that and the changes for PI approved status part of another PR. If the role in Keycloak indicates pi status has been granted, I think it makes sense to rename to `is_pi`

chameleon/ChameleonOIDCAuthBackend.py
```diff
@@ -0,0 +1,35 @@
...
+        return user
+
+    def update_user(self, user, claims):
+        logger.debug('Updating user from keycloak with claims: {0}'.format(claims))
+        user.username = claims.get('preferred_username', '')
+        user.first_name = claims.get('given_name', '')
+        user.last_name = claims.get('family_name', '')
+        user.save()
+
+        try:
```
diurnalist, line 28: Nice catch

chameleon/settings.py
diurnalist: OK, sounds like it's supported then :)

chameleon/settings.py
diurnalist: Ah yes, so the default doesn't work for Keycloak's default, makes sense. 👍

requirements.txt
diurnalist: Adding a constraint to less than the next dot release sounds like a sane plan!

PR #92, opened 2020-06-30
Write changes back to origin Keystone server

diurnalist:
This significantly changes how the sync logic works within Portal when a
user is logging. Currently we sync different properties at different
times:

1. The user's enabled/disabled status is synced upon every login to
  Portal (regardless of how initiated)
2. The user's projects are lazily created/enabled/disabled upon every
   login _initiated via Horizon WebSSO_ (an important distinction)
3. The user's memberships in projects are synced upon every login
   initated _via WebSSO_.
4. A PI's membership in a project is synced upon every login of every
   user associated with that project initiated _via WebSSO_.
5. Project nicknames are synced _when they are updated_ in the Portal
   UI, and are not synced during login.
6. Keystone user passwords are synced upon login (regardless of how
  initiated)

All of those write back to the TACC Keystone "primary" server, and
changes are replicated automatically to UC. In order to turn off
replication, we need to ensure each site is updated separately.

This commit thus adds the following logic:

- When a user is logging in via WebSSO, detect for which region they are
  logging in. This will be passed from Horizon.
- Look up the Keystone URL for that region (configured in Portal) and
  perform any operations directly against that Keystone service.
- Operations that take place "out of band" of a login (e.g., setting a
  project nickname) will update all configured Keystones sequentially.

This should ensure that users are able to access the resources on the
site they are trying to log in to (TACC or UC).

NOTE: currently all of the code assumes that the user's domain in
Keystone will match the service account's domain. This will not be
correct when users are federated (they will be in their own domain), but
all of this code only applies to legacy users anyways, so it's OK.


diurnalist:
@zhenz please leave feedback on the allocations code to make sure I did not mess anything up there. I simplified the admin session generator function a bit is all.

@arocha I did some renaming to stay sane while stepping through all of the cases and to call out when the admin client was in use.

chameleon/views.py
```diff
@@ -32,6 +32,9 @@
 
 @login_required
 def horizon_sso_login(request):
+    services_region = request.GET.get('services_region')
```
diurnalist, line 35: See https://github.com/ChameleonCloud/horizon/pull/17

chameleon/views.py
```diff
@@ -96,26 +104,30 @@ def redirect_to_horizon(request):
...
 
 def sync_user_project_status(ks_admin, user, chameleon_projects):
     ks_user = ks_admin.users.list(name=user.username)
     active_projects_in_tas = chameleon_projects.get('active_projects')
-    ks_projects = []
+
     if ks_user:
         ks_projects = ks_admin.projects.list(user=ks_user[0])
+    else:
+        ks_projects = []
```
xandroc, line 119: ?

chameleon/os_login.py
```diff
@@ -47,14 +47,10 @@ def get_unscoped_token(request):
...
     try:
         logger.info('Synchronizing password to keystone for user: ' + request.POST.get('username'))
-        auth = v3.Password(auth_url=settings.OPENSTACK_KEYSTONE_URL,username=settings.OPENSTACK_SERVICE_USERNAME, password=settings.OPENSTACK_SERVICE_PASSWORD, \
-            project_id=settings.OPENSTACK_SERVICE_PROJECT_ID, project_name='services', user_domain_id="default")
-        sess = session.Session(auth=auth, timeout=5)
-        sess = adapter.Adapter(sess, interface='public', region_name=settings.OPENSTACK_TACC_REGION)
-        ks = ks_client.Client(session=sess, interface='public', region_name=settings.OPENSTACK_TACC_REGION)
-        user = filter(lambda this: this.name==request.POST.get('username'), ks.users.list())
+        ks_admin = get_admin_ks_client(request=request)
+        user = filter(lambda this: this.name==request.POST.get('username'), ks_admin.users.list())
```
xandroc, line 51: This looks cool, but it's expensive (I kno, I kno, it was already there, it was a carry-over from the sync script code ), with `ks_user = ks_admin.users.list(name=username)`  we can let keystone query the db for a specific user instead of asking it so get and return all users only for us to iterate through a list to get the one we need

diurnalist: Cool, it is probably possible to do a direct lookup via username, I will check.

xandroc: Should be able to do it with, `ks_user = ks_admin.users.list(name=username)`

chameleon/views.py
diurnalist: Ah, this must have been part of my "working through the code". Sometimes I move things to where I would expect to find them--I tend to avoid re-declaring variables, but it is a personal preference and syntactically is equivalent.

xandroc: :D

PR #93, opened 2020-07-01
manage project metadata using portal db

zhenz:


zhenz:
@diurnalist @xandroc WIP
Tested the following as a federated user:
- loading dashboard
- loading project page
- view project details
- view admin pages
- recharge allocation
- create allocation
- create project
- approve/reject allocation
- update nickname
- add publications
- view publications
- add/remove users to/from project

TODO:
- update expiring/activating allocation script and test
- test the operations as a non-federated user

zhenz:
@diurnalist @xandroc Done! Please review! Thanks!

util/scripts/sync_tas/sync_project_from_tas.py
```diff
@@ -0,0 +1,235 @@
...
+    
+    for proj_charge_code in tas_projects.keys():
+        if not columns: columns = tas_projects[proj_charge_code].keys()
+        if proj_charge_code in portal_projects:
+            portal_proj_id = portal_projects[proj_charge_code]['id']
+            del portal_projects[proj_charge_code]['id']
+            if tas_projects[proj_charge_code] != portal_projects[proj_charge_code]:
+                proj_values = []
+                for key in tas_projects[proj_charge_code].keys():
+                    proj_values.append(key)
```
diurnalist, line 132: I don't understand, both the key and the value are inserted?

util/project_allocation_mapper.py
```diff
@@ -0,0 +1,509 @@
...
+                # assign role
+                role = 'Standard'
+                if username == pi_username:
+                    role = 'PI'
+                users.append(self.portal_keycloak_user_to_tas_obj(username, to_pytas_model=True, role=role))
+        else:
+            users = tas_project.get_users()
+        return users
+    
+    def get_project(self, project_id):
```
diurnalist, line 234: This function would be where you'd first try to fetch the project from the DB and then fall back to TAS, before giving up entirely.

util/project_allocation_mapper.py
```diff
@@ -0,0 +1,509 @@
...
+            tas_project = tas_proj(project_id)
+            try:
+                extras = ProjectExtras.objects.get(tas_project_id=tas_project.id)
+                tas_project.__dict__['nickname'] = extras.nickname
+            except ProjectExtras.DoesNotExist:
+                project_nickname = None
+            return tas_project
+
+    def allocation_approval(self, data, host):
+        if self.is_from_db:
```
diurnalist, line 254: The code choice here depends on if the _project_ is federated, not the user. Similarly, try to fetch it from portal first--if it is found, you know it's in the "DB" case.

util/project_allocation_mapper.py
```diff
@@ -0,0 +1,509 @@
...
+            return project.chargeCode, project.nickname, project.description
+        else:
+            project_extras = ProjectExtras.objects.filter(tas_project_id=project.id)
+            name = project.chargeCode
+            if project_extras:
+                name = project_extras[0].nickname
+            return project.chargeCode, name, project.description
+        
+    def add_user_to_project(self, tas_project, username):
+        if self.is_from_db:
```
diurnalist, line 300: I think actually most of these calls need to check the existence of the project in Portal before doing anything.

allocations/tasks.py
```diff
@@ -21,22 +21,27 @@ def expire_allocations(balance_service):
...
         except Exception:
-            logger.exception('Error expiring project {}'.format(alloc.project_charge_code))
+            logger.exception('Error expiring project {}'.format(alloc.project.charge_code))
 
     logger.debug('need to expire {} allocations, and {} were actually expired'.format(len(expired_allocations), expired_alloc_count))
 
 def deactivate_multiple_active_allocations_of_projects():
-    for proj in Allocation.objects.order_by().values_list('project_charge_code', flat=True).distinct():
-        project_active_allocations = list(Allocation.objects.filter(status='active', project_charge_code=proj))
+    for proj_id in Allocation.objects.order_by().values_list('project_id', flat=True).distinct():
```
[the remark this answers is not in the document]
zhenz: no. just want to order the outputs by project id, so the records with the same project id clustered together for the following process.

projects/migrations/0008_auto_20200625_0937.py
```diff
@@ -0,0 +1,49 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('projects', '0007_auto_20200212_1429'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Field',
```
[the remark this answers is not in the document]
zhenz: The field is the field of science. I think TAS is using the NSF research area to label projects. https://www.nsf.gov/about/research_areas.jsp#:~:text=NSF%20is%20divided%20into%20the,Sciences%2C%20and%20Education%20and%20Human. For the type, I'm not quite sure, but seems like currently TAS only has two types - Startup and Research. And when creating a project, the PI can't select the type (they can select the field). The type is default to Startup. https://github.com/ChameleonCloud/portal/blob/master/projects/views.py#L481. However, I think the type field is somehow useful in the future, so we can label the project as "Research" or "Education" or "Chameleon Training" etc. Currently, if we ask how many projects are used for classes, it's difficult to answer.

util/scripts/sync_tas/sync_project_from_tas.py
zhenz: that's for update query. "update <table> set <column> = <value>"

util/project_allocation_mapper.py
```diff
@@ -0,0 +1,511 @@
...
+            keycloak_client.update_membership(self.get_attr(tas_project, 'chargeCode'), username, 'add')
+            return tas_project.add_user(username)
+        except:
+            return False
+    
+    def remove_user_from_project(self, tas_project, username):
+        try:
+            keycloak_client = KeycloakClient()
+            keycloak_client.update_membership(self.get_attr(tas_project, 'chargeCode'), username, 'delete')
+            return True
```
diurnalist, line 319:
```suggestion
            return tas_project.remove_user(username)
```

And then `return False` in the except I think should do it. We should ensure we log when the update fails.

PR #94, opened 2020-07-01
Disallow editing user profile when using fed auth

xandroc:
Updating user profile form to disable profile info changes (excluding pi eligibility request) when user authenticated via Keycloak.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,4 @@ package-lock.json
 
 #tmp folder
 */.tmp
+celerybeat-schedule
--- a/tas/forms.py
+++ b/tas/forms.py
@@ -2,7 +2,7 @@ from django import forms
 from pytas.http import TASClient
 import re
 import logging
-
+logger = logging.getLogger(__name__)
 
 ELIGIBLE = 'Eligible'
 INELIGIBLE = 'Ineligible'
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -82,3 +82,4 @@ package-lock.json
 
 #tmp folder
 */.tmp
+celerybeat-schedule
```
diurnalist, line 85: ?

tas/forms.py
```diff
@@ -158,15 +158,16 @@ class UserProfileForm(forms.Form):
...
         super(UserProfileForm, self).__init__(*args, **kwargs)
         self.fields['institutionId'].choices = get_institution_choices()
-
         data = self.data or self.initial
         if (data is not None and 'institutionId' in data and data['institutionId']):
             self.fields['departmentId'].choices = get_department_choices(data['institutionId'])
-
+        if is_federated:
+            for field in self.fields:
+                self.fields[field].widget.attrs['readonly'] = True
```
diurnalist, line 169: Will this disable the PI eligibility request checkbox? As discussed, that's the only bit we need to remain interactive. I am not sure how else the PI eligibility can be requested, unless I have forgotten a detail.

.gitignore
xandroc: it's generated by celery on startup, didn't think we should check it in. Doesn't happen for you?

tas/forms.py
xandroc: No, Request PI Eligibility is not a property on the form by default, instead the ability to make a PI eligibility request (checkbox) is added manually in the template and handled as a special case on POST
https://github.com/ChameleonCloud/portal/blob/pi_eligibility/tas/templates/tas/profile_edit.html#L18
https://github.com/ChameleonCloud/portal/pull/94/files#diff-edcdfb0669e5546028c9cb2a21085ec1L69

.gitignore
diurnalist: I don't know that I've run it with the full celery setup still, so did not notice. That's the type of thing that should be in a gitignore though 👍

tas/forms.py
diurnalist: OK, thanks for the clarification!

PR #95, opened 2020-07-13
Pi Eligibility

xandroc:


tas/views.py
```diff
@@ -45,42 +49,63 @@ def get_departments_json(request):
...
 
-def _create_ticket_for_pi_request(user):
+def _create_ticket_for_pi_request(user, is_federated=False):
     """
     This is a stop-gap solution for https://collab.tacc.utexas.edu/issues/8327.
     """
     rt = rtUtil.DjangoRt()
     subject = "Chameleon PI Eligibility Request: %s" % user['username']
+    if is_federated:
+        problem_description = "This PI Eligibility request can be reviewed at " +\
```
diurnalist, line 59: Nice--so we will still send all the eligibility requests to RT, we will just have to update ticketmasters to know that they must be approved in Portal exclusively? Or to both TAS and Portal?

chameleon/models.py
```diff
@@ -1,17 +1,56 @@
...
 class UserProperties(models.Model):
     user = models.OneToOneField(settings.AUTH_USER_MODEL, primary_key=True)
-    PI_STATUS_CHOICES = (
-        ('INELIGIBLE', 'Ineligible'),
+
+    class Meta:
+        verbose_name = ("User Properties")
+        verbose_name_plural = ("User Properties")
+
+    def pi_eligibility():
```
diurnalist, line 13: Which of these two functions is being used? This one, or the one with the decorator? I like this one more because it's less implicit, but maybe it's unused?

xandroc: Yes, this one is unused, it's an alternative to using the decorator's more implicit design. I can update to use this method instead, should be a quick update and test.

tas/views.py
xandroc: As-is a PI Eligibility request made by federated users doesn't make an equivalent request in TAS so those requests would need to be approved in Portal, the text here should let ticketmasters know which requests should be reviewed in portal

xandroc:
> Nice work! I like it. Any reason we need to update PI status on TAS still for federated users? I should think no.

I don't think so, at least not if we're cutting over to federated auth all-at-once, otherwise some users may see themselves as PI Eligible when using one login method and not the other

PR #96, opened 2020-07-24
Change User Profile data source when using Federated Auth

xandroc:
User profile information is currently pulled from TAS. This should change when using federated auth for login. When using the federated auth flow users should see user profile information retrieved from Keycloak

### DIFF
--- a/tas/views.py
+++ b/tas/views.py
@@ -77,21 +77,21 @@ def profile_edit(request):
     is_federated = request.session.get('is_federated', False)
     kwargs = {'is_federated':is_federated}
     if is_federated:
-        tas_user = {}
-        tas_user['firstName'] = request.user.first_name
-        tas_user['lastName'] = request.user.last_name
-        tas_user['email'] = request.user.email
+        user_info = {}
+        user_info['firstName'] = request.user.first_name
+        user_info['lastName'] = request.user.last_name
+        user_info['email'] = request.user.email
     else:
         tas = TASClient()
-        tas_user = tas.get_user(username=request.user)
+        user_info = tas.get_user(username=request.user)
     if is_federated:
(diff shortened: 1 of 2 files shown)


tas/views.py
```diff
@@ -79,17 +92,22 @@ def profile_edit(request):
     if request.method == 'POST':
         request_pi_eligibility = request.POST.get('request_pi_eligibility')
         form = UserProfileForm(request.POST, initial=tas_user)
+        '''
+            if user is federated only check/update PI Eligibility Requests
+        '''
+        if is_federated and request_pi_eligibility:
```
diurnalist, line 98: I think we probably still want to put this under `if form.is_valid()` - that likely performs CSRF checking, yes?

tas/views.py
```diff
@@ -79,17 +92,22 @@ def profile_edit(request):
...
+            pie_request = PIEligibility()
+            pie_request.requestor = request.user
+            pie_request.save()
+            _create_ticket_for_pi_request(tas_user, is_federated)
```
diurnalist, line 102: What does this function do for federated users? What if they're not in TAS?

tas/views.py
xandroc: I think the CSRF check happens in the middleware, https://docs.djangoproject.com/en/1.11/ref/csrf/#how-to-use-it

tas/views.py
xandroc: This creates a pi eligibility request in django, then creates a ticket for the request. The ticket create function passes in tas_user but that var is pre-populated with non-tas data if the user is federated, could update the name though.

```    if is_federated:
        tas_user = {}
        tas_user['firstName'] = request.user.first_name
        tas_user['lastName'] = request.user.last_name
        tas_user['email'] = request.user.email```

PR #100, opened 2020-08-11
fix sync scripts

zhenz:


util/scripts/sync_tas/sync_allocation_from_tas.py
```diff
@@ -113,19 +114,25 @@ def sync(db, tas_allocs, portal_allocs):
     for tas_alloc in tas_allocs:
         if not columns:
             columns = tas_alloc.keys()
-        # tas projects_for_group gives inconsistent computeUsed
-        compare_tas_alloc = tas_alloc.copy()
-        del compare_tas_alloc['su_used']
+        compare_tas_alloc = {}
```
diurnalist, line 117: I think it may be easier to reason about this by implementing a compare function:

```python
def alloc_equals(portal_alloc, tas_alloc):
    return all(tas_alloc.get(key) == portal_alloc.get(key) for key in ALLOCATION_ID_KEYS)
```

Then you can simply do

```python
for portal_alloc in portal_allocs:
    if alloc_equals(portal_alloc, tas_alloc):
        exist = True

File: util/scripts/sync_tas/sync_allocation_from_tas.py [hunk 2]
```diff
@@ -113,19 +114,25 @@ def sync(db, tas_allocs, portal_allocs):
...
-            if compare_tas_alloc == compare_portal_alloc:
-                exist = True
-                if tas_alloc['su_used'] != portal_alloc['su_used'] and tas_alloc['su_used'] > 0.0:
-                    records_to_update.append((tas_alloc['su_used'], portal_alloc['id']))
-                break
+            compare_portal_alloc = {}
+            for key in portal_alloc.keys():
+                if key in ALLOCATION_ID_KEYS:
+                    compare_portal_alloc[key] = portal_alloc[key]
+                if compare_tas_alloc == compare_portal_alloc:
```
diurnalist, line 127: I think this is too far indented

util/scripts/sync_tas/sync_allocation_from_tas.py
```diff
@@ -113,19 +114,25 @@ def sync(db, tas_allocs, portal_allocs):
...
+                    exist = True
+                    portal_alloc_id = portal_alloc['id']
+                    del portal_alloc['id']
+                    if portal_alloc != tas_alloc:
```
diurnalist, line 131: By this point they have already been checked for a fuzzy "equality" - on what keys can then then differ? I think this could be another function, which might make it more clear:

```python
def alloc_delta(portal_alloc, tas_alloc):
    """Compute the difference between a Portal and a TAS allocation.
    """
    ignored_keys = ALLOCATION_ID_KEYS + ['id']
    return {
        key: tas_alloc.get(key) for key in tas_alloc.keys()
        if portal_alloc.get(key) != tas_alloc.get(key)
    }
```

Then you can check if there is any delta:

```python
delta = alloc_delta(portal_alloc, tas_alloc)
if delta:
    alloc_values = [delta[key] for key in sorted(delta.keys())] + [portal_alloc['id']]
    records_to_update.append(alloc_values)
```

PR #104, opened 2020-08-14
Adding PI Eligiblity sync script, using single logger entry for util …

xandroc:
…module

### DIFF
--- a/chameleon/settings.py
+++ b/chameleon/settings.py
@@ -378,13 +378,9 @@ LOGGING = {
             'handlers': ['console'],
             'level': 'INFO',
         },
-        'util.project_allocation_mapper': {
+        'util': {
             'handlers': ['console'],
-            'level': 'INFO',
-        },
-        'util.keycloak_client': {
-            'handlers': ['console'],
-            'level': 'INFO',
+            'level': 'DEBUG',
         },
     },
 }
--- a/util/scripts/sync_tas/sync_pi_status_from_tas.py
+++ b/util/scripts/sync_tas/sync_pi_status_from_tas.py
@@ -45,7 +45,6 @@ def main(argv=None):
     else:
         updated = import_pi_status(args.dryrun, users, TASClient())
 
-    
     finished = datetime.now()
     if(args.dryrun):
         logger.info('Dry run, no updates made')
@@ -89,10 +88,10 @@ def import_pi_status(dryrun=False, users=[], tas=None, set_eligible=False):
 def set_all_pi_status(dryrun=False, users=[], status=None):
     updated = []
     for index, user in enumerate(users):
+        percent_complete = int(index/float(users.count()) * 100)
+        logger.info('{0} of {1}, {2}% complete, next: {3}'.format(\
+            index, users.count(), percent_complete, user.username))
         if user.pi_eligibility().lower() == status.lower():
-            percent_complete = int(index/float(users.count()) * 100)
-            logger.info('{0} of {1}, {2}% complete, next: {3}'.format(\
-                index, users.count(), percent_complete, user.username))
             logger.info('pi_eligibility {0} already set for user {1}'.format(\
                 status, user.username))
             continue
--- a/util/scripts/sync_tas/sync_pi_status_from_tas.py
+++ b/util/scripts/sync_tas/sync_pi_status_from_tas.py
@@ -95,7 +95,7 @@ def set_all_pi_status(dryrun=False, users=[], status=None):
             logger.info('pi_eligibility {0} already set for user {1}'.format(\
                 status, user.username))
             continue
-        save_pi_status(user, "Eligible", dryrun)
+        save_pi_status(user, status, dryrun)
         updated.append((user.username, "Eligible"))
     return updated
 
@@ -122,5 +122,4 @@ def save_pi_status(user, status, dryrun=False):
 
 if __name__=="__main__":
     main()
-
 
--- a/util/scripts/sync_tas/sync_pi_status_from_tas.py
+++ b/util/scripts/sync_tas/sync_pi_status_from_tas.py
@@ -60,29 +60,25 @@ def main(argv=None):
     delta = finished - started
     logger.info('elapsed seconds = {0}'.format(delta.total_seconds()))
 
-def import_pi_status(dryrun=False, users=[], tas=None, set_eligible=False):
+def import_pi_status(dryrun=False, users=[], tas=None):
     updated = []
-    if set_eligible:
-        for user in users:
-            save_pi_status(user, "Eligible", dryrun)
-            updated.append((user.username, ["Eligible"]))
-    else:
-        for index, user in enumerate(users):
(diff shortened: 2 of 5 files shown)


chameleon/settings.py
```diff
@@ -378,13 +378,9 @@
             'handlers': ['console'],
             'level': 'INFO',
         },
-        'util.project_allocation_mapper': {
```
diurnalist, line 381: 👍

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
+import os
+import sys
+import logging
+import django
+
+# loading up django so we can use models for queries/updates
+sys.path.append('/project') 
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "chameleon.settings")
+django.setup()
```
diurnalist, line 9: What does this do?

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+        'Simulate import, no changes applied', default=False)
+    parser.add_argument('--users', type=str, help=\
+        'Synchronize a specific user or set of users using a csv list of usernames', default=None)
+    parser.add_argument('--seteligible', type=bool, help=\
+        'Used with --users flag to set all provided users as PI Eligible', default=False)
+    args = parser.parse_args(argv[1:])
+
+    started = datetime.now()
+    logger.info('Starting... {0}'.format(started))
+    usermodel = get_user_model()
```
diurnalist, line 36: Ah, it lets you do this, I guess?

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+    if args.users:
+        usernames = args.users.split(',')
+        users = usermodel.objects.filter(username__in=usernames)
+    else:
+        users = usermodel.objects.all()
+
+    if args.users and args.seteligible:
+        updated = set_all_pi_status(args.dryrun, users, "Eligible")
+    else:
+        updated = import_pi_status(args.dryrun, users, TASClient())
```
diurnalist, line 46: The `set_eligible` flag is never used on this function, should that code path inside import_pi_status be removed?

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+
+def import_pi_status(dryrun=False, users=[], tas=None, set_eligible=False):
+    updated = []
+    if set_eligible:
+        for user in users:
+            save_pi_status(user, "Eligible", dryrun)
+            updated.append((user.username, ["Eligible"]))
+    else:
+        for index, user in enumerate(users):
+            percent_complete = int(index/float(users.count()) * 100)
```
diurnalist, line 71: well aren't you fancy

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+            logger.info('{0} of {1}, {2}% complete, next: {3}'.format(\
+                index, users.count(), percent_complete, user.username))
+            if user.pi_eligibility().lower() == 'ineligible': 
+                ''' user is ineligible in portal, let's see if TAS says different '''
+                tas_pi_status = get_tas_pi_status(tas, user.username)
+                if tas_pi_status.lower() != user.pi_eligibility().lower():
```
diurnalist, line 77: Possible this could wind up setting pi eligibility to an empty string? What possible values are there for this? I thought it was just "eligible" and "ineligible".

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+    updated = []
+    for index, user in enumerate(users):
+        percent_complete = int(index/float(users.count()) * 100)
+        logger.info('{0} of {1}, {2}% complete, next: {3}'.format(\
+            index, users.count(), percent_complete, user.username))
+        if user.pi_eligibility().lower() == status.lower():
+            logger.info('pi_eligibility {0} already set for user {1}'.format(\
+                status, user.username))
+            continue
+        save_pi_status(user, "Eligible", dryrun)
```
diurnalist, line 98:
```suggestion
        save_pi_status(user, status, dryrun)
```

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+        updated.append((user.username, "Eligible"))
```
diurnalist, line 99:
```suggestion
        updated.append((user.username, status))
```

util/scripts/sync_tas/sync_pi_status_from_tas.py
```diff
@@ -0,0 +1,126 @@
...
+        return tas.get_user(username=username)['piEligibility']
+    except:
+        logger.error('Error getting pi status for user: {0}'.format(username), sys.exc_info()[0])
+        return 'Ineligible'
+
+def save_pi_status(user, status, dryrun=False):
+    pie = PIEligibility()
+    pie.requestor = user
+    pie.status = status.upper()
+    pie.review_summary = 'Imported From TAS'
```
diurnalist, line 113: :+1:

util/scripts/sync_tas/sync_pi_status_from_tas.py
xandroc: this loads django so we can run this as a script independent of the django instance managed by uwsgi (so we don't have to create a url to reach the logic or worry about timeouts) and still use the django models api to read/write to the db

util/scripts/sync_tas/sync_pi_status_from_tas.py
xandroc: also `requested`,
Could, but the empty string would have to come from TAS I think, the data looked clean.

util/scripts/sync_tas/sync_pi_status_from_tas.py
xandroc: it does, as long as i spool up django and point it to the settings file

util/scripts/sync_tas/sync_pi_status_from_tas.py
xandroc: i think I see what you mean, I think the field is pre-populated with Ineligible by default for all users, if it changed it would be to requested or eligible, but it could be empty if it was set that way by the api and TAS didn't kick it back. I did run this locally and in dev with no empties tho

util/scripts/sync_tas/sync_pi_status_from_tas.py
xandroc: yes, good catch, thanks!

PR #107, opened 2020-08-27
Sort project/allocations by requested date, most recent first

xandroc:


util/project_allocation_mapper.py
```diff
@@ -91,7 +91,18 @@ def get_all_projects(self):
...
+        values = projects.values()
+        values.sort(reverse=True,key=self.sort_by_allocation_request_date)
+        return values
+
+    '''
+    Sort by most recent allocation request
+    '''
+    def sort_by_allocation_request_date(self, el):
+        if el['allocations']:
+            return datetime.strptime(el['allocations'][0]['dateRequested'], '%Y-%m-%dT%H:%M:%SZ')
```
diurnalist, line 103: Is the allocation in position 0 always going to be the most recent? It might make sense to get the max `dateRequested`.

Also, what if `dateRequested` is not on the allocation? I think this would throw an error in that case, which can probably break a lot of pages. Perhaps defaulting to datetime.min and logging an error is best.

xandroc: good catch, yes it looks like the last item in the allocations list is the most recent. I think you're right on `dateRequested` too, if it happens to be missing we'll probably get an error, i'll make updates and resubmit, thanks!

PR #111, opened 2020-09-03
rochaa/memcached

xandroc:
adding memcached option, also setting django-filer version to one supported by python 2, recent changes made python3.5 minimum for the latest version of django-filer, https://pypi.org/project/django-filer/

### DIFF
--- a/chameleon/settings.py
+++ b/chameleon/settings.py
@@ -719,7 +719,7 @@ CELERY_BEAT_SCHEDULE = {
 CSP_FRAME_ANCESTORS = "'self'"  # Similar to X-Frame-Options: SAMEORIGIN
 CSP_SCRIPT_SRC = ["'self'", 'https://www.google-analytics.com/', "'unsafe-inline'"]
 CSP_FONT_SRC = ["'self'", 'https://fonts.gstatic.com/']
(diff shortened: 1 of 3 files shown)


chameleon/settings.py
```diff
@@ -719,9 +719,17 @@
 CSP_FRAME_ANCESTORS = "'self'"  # Similar to X-Frame-Options: SAMEORIGIN
 CSP_SCRIPT_SRC = ["'self'", 'https://www.google-analytics.com/', "'unsafe-inline'"]
 CSP_FONT_SRC = ["'self'", 'https://fonts.gstatic.com/']
+CSP_IMG_SRC = ["'self'", '*.googleusercontent.com', "'unsafe-inline'"]
```
diurnalist, line 722: I'm not sure that we need `unsafe-inline` here, I'm not sure how that even applies to img elements in CSP actually 🤔

upper-constraints.txt
```diff
@@ -8,6 +8,7 @@ django-bootstrap3~=11.0
 django-ckeditor~=5.3.0
 django-classy-tags~=0.8.0
 django-cms~=3.6.0
+django-filer<2.0.0
```
diurnalist, line 11: This is already in master I think

diurnalist (approved):
LGTM! Think the img-src declaration could be simpler

PR #113, opened 2020-09-09
fix issue of listing random users in a project

zhenz:
When we get a project from portal, although we switch the project as TAS Project object, we keep the portal project id. But this will mess up getting members of a project https://github.com/ChameleonCloud/portal/blob/master/util/project_allocation_mapper.py#L285-L287.

### DIFF
--- a/util/project_allocation_mapper.py
+++ b/util/project_allocation_mapper.py
@@ -282,6 +282,10 @@ class ProjectAllocationMapper:
         # we have stopped writing any data to TAS for projects/allocations/
         # memberships.
         usernames = [u.username for u in users]
+        # The project id might be portal id.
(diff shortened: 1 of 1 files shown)


diurnalist (changes requested):
Pain... good catch. But in how many places are we calling get_project? The other way is to perform this fix in get_project_members... look up the right TAS project immediately before querying TAS. I think that could be a bit more explicit (though obviously, not having it happen earlier in the chain bit us here.) I think I actually prefer that, since overriding the id could have other consequences when we are saving this back to the DB.

diurnalist (approved):
If set_attr in fact returns the project again, then :+1:
Thanks! 🎖️

PR #115, opened 2020-09-25
rochaa/python3

xandroc:
A lot of changes in here, some things to focus on might be requirements.txt and upper-constraints.txt, some of those dependencies may need to be tightened up.


chameleon/celery.py
```diff
@@ -15,4 +15,4 @@
 
 @app.task(bind=True)
 def debug_task(self):
-    print('Request: {0!r}'.format(self.request))
+    print(('Request: {0!r}'.format(self.request)))
```
diurnalist, line 18: Weird that `2to3` did this

chameleon/keystone_auth.py
```diff
@@ -162,7 +162,7 @@ def regenerate_tokens(request, password):
     email = request.user.email
     tokens = {}
 
-    for region in settings.OPENSTACK_AUTH_REGIONS.keys():
+    for region in list(settings.OPENSTACK_AUTH_REGIONS.keys()):
```
diurnalist, line 165: This `list()` optimization is something that `2to3` does just to be safe, but in most cases it is not necessary, as in this case. Fine to leave it in though, as it doesn't hurt anything really for us.

g5k_discovery/tests/django/view_test.py
```diff
@@ -1,32 +1,32 @@
...
+#         mock_render_to_response = mock.MagicMock()
+#         with mock.patch.multiple('g5k_discovery.views',
+#             render_to_response=mock_render_to_response,
+#             RequestContext=mock.MagicMock(),
+#             login_required=lambda x: x):
+#             from g5k_discovery.views import g5k_html
+#             mock_request = mock.Mock()
+#             g5k_html(mock_request, 'site')
+#             _, args, _ = mock_render_to_response.mock_calls[0]
+#             self.assertEquals(args[0], 'g5k_discovery/site.html', 'Wrong template')
```
diurnalist, line 32: If this test isn't useful anymore, we should consider deleting it 💥

user_news/forms.py
```diff
@@ -16,7 +15,7 @@ class OutageForm(ModelForm):
     class Meta:
         model = Outage
         exclude = ['reminder_sent']
-	fields = '__all__'
+        # fields = '__all__'
```
diurnalist, line 18: Why is this one commented out?

diurnalist (approved):
This looks good to me and is about what I'd expect from such a change. Let's get to testing! ⭐

user_news/forms.py
xandroc: setting exclude fields automatically adds all fields except the one in the exclude list, i should probably just remove instead of of commenting it out. The best way to set these is probably to explicitly add all the fields we want to include so we don't inadvertently expose fields that shouldn't go to the front end, but that's probably out of scope for this change

g5k_discovery/tests/django/view_test.py
xandroc: rgr, will do

PR #116, opened 2020-10-02
casting allocation id to int to avoid type comparison exception

zhenz:


xandroc:
This looks good, LGTM! thank you!!

PR #117, opened 2020-10-05
save all projects/allocations to portal and keycloak

zhenz:
Saving projects to TAS will not create project in keycloak, and thus
users will miss their newly created project in the dashboard.


diurnalist (approved):
The keycloak-tas-sync will actually ensure that Keycloak has the group (and users) eventually, but this is a move forward, we should over time just remove all mention of TAS.

PR #118, opened 2020-10-06
use portal for project fetching and approval

zhenz:
Since all project and allocation requests will not go to TAS, need to
fetch and approve projects via portal only.


diurnalist (approved):
Perfect!

PR #121, opened 2020-10-08
lazy add user when log in; use user portal id

zhenz:
1) Users are not added to Keycloak if the user is not part of any active
allocations in TAS, though we have allocations approved for them in portal
DB. Therefore, we add TAS-authenticated users to Keycload when they log in.

2) Projects and allocations are created directly to portal DB. When legacy
users create new allocations, the user id is their TAS id. When using TAS
id as the requestor id and saving the new allocation to the portal DB, it
will cause the foreign key conflict. Therefore, we get the portal user id
as the requestor id.


zhenz:
@diurnalist fixed two things in this PR. the second problem is reported by several users.

util/keycloak_client.py
```diff
@@ -110,3 +134,34 @@ def create_project(self, charge_code, pi_username):
         keycloakproject.create(charge_code)
 
         self.update_membership(charge_code, pi_username, 'add')
+
+    def create_user(self, username, first_name=None, last_name=None, email=None,
+                    affiliation_title=None, affiliation_department=None,
+                    affiliation_institution=None, country=None, citizenship=None, join_date=None):
+        try:
+            keycloakuser = self._users_admin()
+            keycloakuser.create(
```
diurnalist, line 143: Does this not return the user? If it returns the user (with ID), you don't need the following lookup.

diurnalist (approved):
This looks good to me! The other fix also makes sense. Nice commit message!

util/keycloak_client.py
zhenz: it returns an empty string.

PR #122, opened 2020-10-15
improve admin projects/allocations loading

zhenz:
Loading all projects/allocations is slow in admin page. The cause of the
slowness is due to the django model foreign key lazy load. To improve the
performace, django "select_related" is used. The django "select_related"
is a performance booster which results in a single more complex query but
means later use of foreign-key relationships won’t require database queries.

Read https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-related
for more information.

### DIFF
--- a/util/project_allocation_mapper.py
+++ b/util/project_allocation_mapper.py
@@ -67,7 +67,7 @@ class ProjectAllocationMapper:
     def _get_project_allocations(self, project, fetch_balance=True):
         balance_service = BalanceServiceClient()
         reformated_allocations = []
-        for alloc in portal_alloc.objects.filter(project_id=project.id):
+        for alloc in project.allocations.select_related('project', 'requestor', 'reviewer').all():
             if fetch_balance and alloc.status == 'active':
                 balance = balance_service.call(project.charge_code)
                 if balance:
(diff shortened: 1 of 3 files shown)


util/project_allocation_mapper.py
```diff
@@ -67,7 +67,7 @@ def _send_allocation_decision_notification(self, charge_code, requestor_id, stat
     def _get_project_allocations(self, project, fetch_balance=True):
         balance_service = BalanceServiceClient()
         reformated_allocations = []
-        for alloc in portal_alloc.objects.filter(project_id=project.id):
+        for alloc in portal_alloc.objects.select_related('project', 'requestor', 'reviewer').filter(project_id=project.id):
```
diurnalist, line 70: If you've pre-queried for allocations by this point, could you also do `project.allocations` instead of making a new query?

util/project_allocation_mapper.py
```diff
@@ -99,9 +99,17 @@ def _get_user_from_portal_db(self, username):
 
     def get_all_projects(self):
         projects = {}
-        for proj in portal_proj.objects.all():
-            proj = self.portal_to_tas_proj_obj(proj, fetch_balance=False)
-            projects[proj['chargeCode']] = proj
+        project_pi_info = {}
+        for proj in portal_proj.objects.select_related('pi').all():
```
diurnalist, line 103: As an alternative, what about selecting all the related things at this level?

```
portal_proj.objects.select_related('pi', 'type', 'field', 'allocations', 'allocations__requestor').all()
```

I'm not sure if that even works, specifically the `__requestor` accessor. But it may compress everything to a single SELECT w/ multiple joins.

diurnalist (approved):
Nice commit message :+1: - I can understand how this improves things, but do you have any numbers for how much of an improvement? Does it reduce the amount of queries significantly in practice, and how much speed-up does it yield?

Left some comment on the allocations fetching.

Nice sleuthing :) 🕵️‍♀️

util/project_allocation_mapper.py
zhenz: yes. could do that. will fix.

util/project_allocation_mapper.py
zhenz: Not sure about "allocations" and "allocations__requestor". but since we only use "pi", adding more things and making it a more complex query will increase the exec time.

PR #123, opened 2020-10-19
Set flag on Keycloak groups w/ active alloc

diurnalist:
Beyond just the allocation balance, we sometimes need to know the
standing of a project. There is no way to enable/disable a group in
Keycloak, so instead we must add an additional piece of metadata and
keep it in sync w/ the allocation. We expire and activate allocations
via this task mechanism, so that's where this logic lives.


zhenz (approved):
LGTM!

PR #126, opened 2020-10-21
show user metadata in pi eligibility review page and profile page

zhenz:
Portal doesn't store all metadata of a user. To provide the user
information for pi eligibility reviewers and to show profile in the user
dashboard, we fetch user metadata from Keycloak.

### DIFF
--- a/chameleon/admin.py
+++ b/chameleon/admin.py
@@ -3,6 +3,8 @@ from functools import wraps
 from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.auth import get_user_model
 from chameleon.models import PIEligibility
+from util.keycloak_client import KeycloakClient
+import re
 
 def add_method(cls):
     def decorator(func):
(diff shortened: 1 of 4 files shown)


zhenz:
@diurnalist I have updated the pi eligibility review page and user profile page showing user metadata in this PR. I'll continue working on the edit profile page and maybe create a separate PR.

chameleon/admin.py
```diff
@@ -21,12 +23,35 @@ def pi_eligibility(self):
...
+    fields = ('requestor','request_date','user_metadata','status','review_date','reviewer','review_summary')
     ordering = ['-status','-request_date','-review_date']
 
     list_display = ('requestor','status','request_date')
     list_filter = ('status',)
     search_fields = ['requestor__username']
 
+    def user_metadata(self, obj):
+        keycloak_client = KeycloakClient()
+        keycloak_user = keycloak_client.get_keycloak_user_by_username(obj.requestor.username)
```
diurnalist, line 36: Keycloak's user attribute structure is a bit wacky. What if we normalize it in this `get_keycloak_user_by_username` function? You could iterate over the attributes, find all the ones that are list of size 1, and flatten them so it's just a key/value pair, as we'd expect. If for some reason the list is size > 1 (multivalued attribute I guess, which we don't utilize at the moment), you could leave it as a list, which I think would be the expected behavior.

If you normalize here, we don't need to have that extractor function, which seems to simplify things a lot.

diurnalist (changes requested):
Little  suggestion, otherwise this looks very nice. I'm guessing that the user_metadata function is only called when rendering the detail view? As in, when you load the list of pi eligibility requests, it doesn't request every user from Keycloak, does it?

zhenz:
> Little suggestion, otherwise this looks very nice. I'm guessing that the user_metadata function is only called when rendering the detail view? As in, when you load the list of pi eligibility requests, it doesn't request every user from Keycloak, does it?

No. only when rendering the detail page.

util/project_allocation_mapper.py
```diff
@@ -570,8 +571,17 @@ def portal_user_to_tas_obj(self, user, role='Standard'):
                     'department': None,
                     'institution': None,
                     'role': role}
-        return tas_user
 
+        keycloak_client = KeycloakClient()
+        keycloak_user = keycloak_client.get_keycloak_user_by_username(user.username)
+
+        if keycloak_user:
+            for key, val in keycloak_user['attributes'].items():
```
diurnalist, line 579: In cases like this, where we're mapping between two data types, I'd actually prefer to be explicit:

```python
attrs = keycloak_user['attributes']
tas_user.update({
  'institution': attrs.get('affiliationInstitution'),
  'department': attrs.get('affiliationDepartment'),
  # ... and so on
})
```

The main benefit is that, if I'm coming along later and I want to change the name "affiliationInstitution" to "affiliationUniversity" or something, it's easier for me to find all the places we use that key, because I can simply search for it. Search visibility is a good property when you don't have strict types to fall back on, which normally are what you leverage in such a refactor.

diurnalist (approved):
Looks great!

PR #127, opened 2020-10-26
allow edit profile for users

zhenz:


tas/views.py
```diff
@@ -43,73 +42,31 @@ def get_departments_json(request):
...
-            pie_request.requestor = request.user
-            pie_request.save()
-            _create_ticket_for_pi_request(user_info, is_federated)
-            messages.success(request, 'Your profile has been updated!')
-            return HttpResponseRedirect(reverse('tas:profile'))
-        '''
-            If not federated validate form and update TAS profile info
-        '''
+        if request_pi_eligibility:
+            form.pi_eligible()
```
diurnalist, line 54: This seems like it would be better as an argument passed to the `UserProfileForm` constructor. As a function, it is a good example of a [side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)), a pattern which comes with some significant downsides.

tas/views.py
```diff
@@ -43,73 +42,31 @@ def get_departments_json(request):
...
-            else:
-                data['piEligibility'] = user_info['piEligibility']
-            data['source'] = 'Chameleon'
-            tas.save_user(user_info['id'], data)
+            mapper.update_user_profile(user_info, data, request_pi_eligibility)
             messages.success(request, 'Your profile has been updated!')
             return HttpResponseRedirect(reverse('tas:profile'))
     else:
-        form = UserProfileForm(initial=user_info,**kwargs)
+        form = UserProfileForm(initial=user_info)
```
diurnalist, line 62: similar remark here, if it's added as a constructor argument, it is a bit easier to follow.

tas/views.py
```diff
@@ -43,73 +42,31 @@ def get_departments_json(request):
...
+        if user_info['piEligibility'].upper() == 'ELIGIBLE':
+            form.pi_eligible()
 
     context = {
         'form': form,
         'user': user_info,
-        'piEligibility': piEligibility,
+        'piEligibility': user_info['piEligibility'],
```
diurnalist, line 69: Watch out for the letter case here... I hit a bug with this at some point, where the template was expecting it to look like `"ELIGIBLE"` but it was actually `"Eligible"`.

util/project_allocation_mapper.py
```diff
@@ -97,6 +99,22 @@ def _get_user_from_portal_db(self, username):
             logger.error('Could not find user %s in DB', username)
             return None
 
+    def _create_ticket_for_pi_request(self, user):
+        """
+        This is a stop-gap solution for https://collab.tacc.utexas.edu/issues/8327.
```
diurnalist, line 104: Heh, this ticket is 5 years old. Long live the stop-gap!

util/project_allocation_mapper.py
```diff
@@ -237,6 +255,26 @@ def update_project_nickname(project_id, project_charge_code, nickname):
...
+                pie_request = PIEligibility()
+                pie_request.requestor_id = user['id']
+                pie_request.save()
+            else:
+                self.tas.save_user(user['id'], {'piEligibility': 'Requested'})
+            self._create_ticket_for_pi_request(user)
+
+        keycloak_client = KeycloakClient()
+        keycloak_client.update_user(user['username'],
+                                    affiliation_title=new_profile['title'] if 'title' in new_profile else None,
```
diurnalist, line 271:
```suggestion
                                    affiliation_title=new_profile.get('title'),
```
I think that is equivalent?

diurnalist (approved):
Looks good :) -- given that citizenship is readonly, maybe it makes sense to not even display that as a form field. I just looked through Django docs though and I think this is suprisingly difficult to achieve without rendering the entire form manually, so maybe not worth it.

PR #130, opened 2020-11-06
Cinder migration

jakecoll:


chameleon/tasks.py
```diff
@@ -3,18 +3,19 @@
...
+from .keystone_auth import admin_session, admin_ks_client, get_user, unscoped_user_session, project_scoped_session
 
 LOG = get_task_logger(__name__)
 
 
 class MigrationError(Exception):
     def __init__(self, messages=[]):
         self.messages = messages
-        super(Exception, self).__init__(messages)
+        super(MigrationError, self).__init__(messages)
```
diurnalist, line 18: d'oh! Good catch.

chameleon/tasks.py
```diff
@@ -131,6 +135,43 @@ def _migrate_project(region, username=None, charge_code=None, access_token=None)
...
+        )]
+
+        num_volumes = len(volumes_to_migrate) 
+        yield progress, f'There are {num_volumes} volumes to process.'
+
+        for index, volume in enumerate(volumes_to_migrate):
+
+            if volume.status == 'available':
+                transfer = admin_cinder.transfers.create(volume.id)
+                user_cinder.transfers.accept(transfer.id, transfer.auth_key)
```
diurnalist, line 165: Does it not work to use the admin cinder client for this? That could simplify this.

chameleon/tasks.py
```diff
@@ -131,6 +135,43 @@ def _migrate_project(region, username=None, charge_code=None, access_token=None)
...
+            else:
+                yield progress, (
+                    f'Volume {volume.name} is not available to transfer. '
+                    f'Please detach volume from any instances and re-run '
+                    f'migration.'
+                )
+
+            yield progress, f'Processed {index + 1} of {num_volumes} volumes.'
```
diurnalist, line 173: `progress` is never incremented--incrementing progress is how the progress bar moves in the UI. Incrementing progress is however tricky w/ the current code, which assumes the only thing being migrated are disk images, and so when 100% of the disk images are migrated, the progress is 100% (1.0).

diurnalist (changes requested):
Clean and simple :+1:

We should be adjusting progress as we go to give the user a bit of feedback here.
Also, I'm not sure the user session is a hard requirement when performing the transfer accept, maybe it would work with the admin session too?

chameleon/tasks.py
jakecoll: Unfortunately, no. The cinder client only accepts a transfer id and auth key so it will transfer the volume to the authenticating project (i.e. `openstack` rather than user's federated project).

chameleon/tasks.py
jakecoll: Yea, I can fix this up.

jakecoll: Updated the code. Looks to be working properly.

diurnalist (approved):
Nice changes! This looks good to go 🚢

PR #131, opened 2020-11-09
Add html_message kwarg to outage reminder mail.

jakecoll:
Outage reminder emails were being sent without html formatting.
This was because only plain_text was being passed to django's
send_mail function. To preserve html formatting for the email
message the message body should be passed to html_message kwarg.


diurnalist (approved):
Ah! Thanks

PR #135, opened 2020-12-07
sort order of allocations by request date.

msherman64:
separated method for getting sortable key from allocation
used for sorting both allocations and projects

Extends fix from https://github.com/ChameleonCloud/portal/pull/107

### DIFF
--- a/util/project_allocation_mapper.py
+++ b/util/project_allocation_mapper.py
@@ -88,7 +88,7 @@ class ProjectAllocationMapper:
                 else:
                     logger.warning('Couldn\'t get balance. Balance service might be down.')
             reformated_allocations.append(self.portal_to_tas_alloc_obj(alloc))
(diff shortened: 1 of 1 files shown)


util/project_allocation_mapper.py
```diff
@@ -129,19 +129,22 @@ def get_all_projects(self):
                 projects[proj['chargeCode']] = proj
             projects[proj['chargeCode']]['allocations'].append(self.portal_to_tas_alloc_obj(alloc))
         return sorted(list(projects.values()), reverse=True, key=self.sort_by_allocation_request_date)
```
diurnalist, line 131: Looking at this code... I have no idea how this is currently working. This will return the same key for all projects :/

diurnalist (approved):
Looks good to me! :shipit:

PR #136, opened 2020-12-22
pin referenceapi to latest version from jenkins

msherman64:
pinned the referenceapi version to the current jenkins build.
the tag "latest" does not exist in the registry.

`https://docker.chameleoncloud.org/v2/referenceapi/tags/list?`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
     depends_on:
       - mysql
   referenceapi:
-    image: docker.chameleoncloud.org/referenceapi:latest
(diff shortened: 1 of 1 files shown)


diurnalist:
I'm actually unsure of how/where the referenceapi image is built. There was some work on it recently but I don't know how finished that is. In any event, I think it's better to fix it there by ensuring we publish `latest` as part of the Jenkins pipeline. https://github.com/ChameleonCloud/reference-api/blob/master/Makefile#L14

diurnalist (changes requested):
Let's fix this by fixing the publish part

PR #137, opened 2020-12-23
optimize query and sort allocations

msherman64:


util/project_allocation_mapper.py
```diff
@@ -113,20 +113,21 @@ def _create_ticket_for_pi_request(self, user):
...
-                                               fetch_balance=False,
-                                               pi_info=project_pi_info[alloc.project.charge_code])
-            if proj['chargeCode'] not in projects:
-                projects[proj['chargeCode']] = proj
-            projects[proj['chargeCode']]['allocations'].append(self.portal_to_tas_alloc_obj(alloc))
-        return sorted(list(projects.values()), reverse=True, key=self.sort_by_allocation_request_date)
+        #each project has many allocations; each allocation has a 'date_requested'
+        #for each project, get the most recent date among its allocations, then annotate with 'newest_request'
+        #sort list of projects by 'newest_request'
+        from django.db.models import Max
```
diurnalist, line 126: We usually lift such imports to the top of the file, unless there is a good reason for them to be lazy-evaluated (as in, they are optional, or perhaps there is a weird import loop for some reason.)

util/project_allocation_mapper.py
```diff
@@ -113,20 +113,21 @@ def _create_ticket_for_pi_request(self, user):
                                  requestor = "us@tacc.utexas.edu")
         rt.createTicket(ticket)
 
+    '''
+    get all projects
+    returns list of tas_proj_obj, sorted by most recent allocation
+    allocations in each object are sorted newest to oldest
+    
+    '''
     def get_all_projects(self):
```
diurnalist, line 122: We've been slowly aligning on using [Google's docstring style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) for code documentation like this. In that model, this would look a bit like this (note placement directly under `def`):

```suggestion
    def get_all_projects(self):
        """Get all projects and their allocations for all users.
        
        Returns:
            List[pytas.Project]: a list of project TAS representations,
                sorted by most recent allocation. Allocations in each
                object are sorted newest to oldest.
```

util/project_allocation_mapper.py
```diff
@@ -113,20 +113,21 @@ def _create_ticket_for_pi_request(self, user):
...
-                                               pi_info=project_pi_info[alloc.project.charge_code])
-            if proj['chargeCode'] not in projects:
-                projects[proj['chargeCode']] = proj
-            projects[proj['chargeCode']]['allocations'].append(self.portal_to_tas_alloc_obj(alloc))
-        return sorted(list(projects.values()), reverse=True, key=self.sort_by_allocation_request_date)
+        #each project has many allocations; each allocation has a 'date_requested'
+        #for each project, get the most recent date among its allocations, then annotate with 'newest_request'
+        #sort list of projects by 'newest_request'
+        from django.db.models import Max
+        sorted_projects = portal_proj.objects.annotate(newest_request=Max('allocations__date_requested')).order_by('newest_request').reverse().select_related('pi')
```
diurnalist, line 127: You might be able to sneak in a `.prefetch_related('allocations')` to really speed this up. Not sure how Django will treat the annotation, which asks for the related allocations. There is some way to turn on logging such that you can see the precise SQL queries Django is issuing, which could prove the value. You could also throw it on and see if it likes it. This is used elsewhere to reduce extra queries in Trovi: https://github.com/ChameleonCloud/portal/blob/0384d1247fc09f712fda7b15de87e3be2de5d1fd/sharing_portal/views.py#L164-L175

`prefetch_related` is like `select_related` but it can handle ManyToMany. It _may_ work to just pass 'allocations' into the `select_related`, I don't know Django well enough to be sure.

diurnalist (approved):
This looks like it's going to really help :) -- how much faster does it feel to you when running locally? Left some minor comments.

util/project_allocation_mapper.py
msherman64: real performance improvements need more refactoring
1. no optimization ~7s
2. select_related('type','pi','field') 3.8-4.5s
3. prefetch_related('allocations) 4-4.5s
4. both: 4-4.5s

new view with no frontend, simple conversion from model ->dict
1. json.dumps(portal_proj.objects.all() ) 0.4s
2. json.dumps(portal_alloc.objects.all() ) 0.5s

Easiest win is just restricting the filterset by default, make a new view for:
"projects with at least one pending allocation"

diurnalist (approved):
Great stuff! 🚢

PR #138, opened 2020-12-23
add progress bar for SU usage

msherman64:


allocations/templates/allocations/index.html
```diff
@@ -119,12 +119,12 @@ <h3>Allocations</h3>
                                     <th width="20%">Actions</th>
                                 </tr>
                             </thead>
-                            <tbody>
-                                <tr ng-repeat="allocation in project.allocations" ng-show="!allocation.doNotShow">
+                            <tbody ng-repeat="allocation in project.allocations" ng-show="!allocation.doNotShow">
```
diurnalist, line 122: Huh, I didn't know it was legal in HTML to have multiple `tbody`s. But given the amount of things that are legal in HTML, nothing surprises me anymore.

allocations/templates/allocations/index.html
```diff
@@ -159,11 +159,23 @@ <h4><span class="fa fa-spinner fa-spin"></span><span ng-show="isLoading('approve
...
+                                        <div ng-show="!isEmpty(getMessages('rejectAllocation', allocation))">
                                             <alert ng-repeat="msg in getMessages('rejectAllocation', allocation)" type="[[msg.type]]">[[msg.body]]</alert>
                                         </div>
                                     </td>
                                 </tr>
+                                <tr>
+                                    <td colspan="3">
+                                        <div class="progress">
+                                            <div class="progress-bar" role="progressbar" 
+                                            aria-valuenow=[[100 * allocation.computeAllocated / allocation.computeRequested]]
```
diurnalist, line 171: I think what we want here is `computeUsed / computeAllocated`.

`computeRequested`: how many SUs the user asked for in their request (defaults to 20k, I don't think users can even change this at request time currently.)
`computeAllocated`: how many SUs were ultimately allocated to them. Normally this == `computeRequested`
`computeUsed`: how many SUs have been used so far. This includes pending leases to avoid you from reserving tons of stuff and not having the credit to back it up :)
`computeBalance`: how many SUs are left. I have no idea why we store this separately, it is the value of allocated - used.

allocations/templates/allocations/user-projects.html
```diff
@@ -167,6 +167,18 @@ <h3>Allocations</h3>
                                             </div>
                                         </td>
                                     </tr>
+                                    <tr>
+                                        <td colspan="3">
+                                            <div class="progress">
+                                                <div class="progress-bar" role="progressbar" 
+                                                aria-valuenow=[[100 * allocation.computeAllocated / allocation.computeRequested]]
```
diurnalist, line 174: Same comment as above.

diurnalist (changes requested):
Code looks good, I like that this could be done just in the view templates w/o any heavy template logic.

I think we need to change the value shown and how it's calculated. Otherwise good to go!

allocations/templates/allocations/index.html
msherman64: I didn't know I couldn't do that 👍 
But the internet says it's ok https://stackoverflow.com/a/3076721

diurnalist (approved):
Very nice, I'm surprised there was no `computeUsed` on the transient model O_O

PR #141, opened 2021-01-08
Migrations 01 07 2021

msherman64:
This commit adds a makefile command to generate the migrations files.

It also contains the migrations from the latest series of commits, due to an addition to transientmodels.


diurnalist:
There are a lot more model changes in here than I thought. Possible to scope the migrations to just your app? I am not sure why a lot of the others are even in there, would have to look at each app's history. Something smells fishy...

msherman64:
Good catch. Restricted it to "allocations", and updated the docs.

diurnalist (approved):
LGTM

PR #146, opened 2021-01-15
Feature/poetry resolve deps

msherman64:
This pull request demonstrates the use of pyproject.toml for linter configuration, as well as for project dependency management.

The use of Poetry primarily brings good dependency constraint resolution, and automatic venv creation during development.


diurnalist:
Should we ignore the get_poetry file since it fails lint? Or, perhaps pull it down dynamically and not vendor it?

msherman64:
Definitely an option. I wasn't sure how you felt about piping curl -> shell from a security standpoint, though I'm not sure there's much difference here.

Technically you can pip install poetry, but I'm not sure of the tradeoffs there.

diurnalist (approved):

diurnalist:
TBH I'm not sure of what the consequences of RCE within a Docker build context are. I should effectively be building everything in a container build context sandbox but I couldn't find any information about security issues there.

This is one of those things where I don't have strong feelings. Vendoring has the downside of the vendored install method getting out of date at some point. Not vendoring has the downside of the possible security issue you mention (which would require the poetry dist server or GitHub repo to be compromised), but also can be susceptible to the same issue as vendoring... the get_poetry.py script might move locations at some point and break things. Let's just keep with vendoring the file for now.

Dockerfile
```diff
@@ -1,26 +1,50 @@
...
-COPY upper-constraints.txt /setup/upper-constraints.txt
-RUN pip install --upgrade pip \
-    && pip install -r /setup/requirements.txt -c /setup/upper-constraints.txt
+WORKDIR /setup
+
+# Use pip to install poetry. We don't use virtualenvs in the build context.
+# Therefore, the vendored install provides no additional isolation.
+RUN pip install \
+  poetry~=1.1 \
+  uWSGI~=2.0
```
msherman64, line 35: Can this be in requirements / pyproject.toml?

Dockerfile
```diff
@@ -1,26 +1,50 @@
...
-  && pip install uwsgi \
-  && npm install -g yuglify
+# Set shell to use for run commands
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Install repos for node6.x. 
+# WARNING: EOL on 2019-04-30
+# https://github.com/nodejs/Release#end-of-life-releases
+ARG NODE_VER=lts
+RUN curl -sL https://deb.nodesource.com/setup_${NODE_VER}.x | bash -
```
msherman64, line 11: we were using "lts" in the client build script, but "6.x" here, which became very outdated.

Dockerfile
```diff
@@ -1,26 +1,50 @@
-FROM python:3.7.9-stretch
+ARG PY_IMG_TAG=3.7.9-stretch
```
msherman64, line 1: Having the build-arg specified should allow us to do things like run tests across a range of python/node versions.

msherman64:
misspell identified an upstream issue: https://github.com/openstack/os-service-types/blob/master/setup.cfg#L3

diurnalist (approved):
🚢

PR #149, opened 2021-01-21
Make docker for local dev more robust

msherman64:
Added some health checks to improve the robustness of the local development environment.

There was often a race between the mysql container and celery.

Health check added to portal container, just to ensure the service is listening.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,13 @@
-version: '3.5'
+version: "3.5"
 
 services:
   portal:
     image: ${DOCKER_IMAGE_LATEST}
+    restart: on-failure
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8888"]
+      timeout: 10s
+      retries: 10
     env_file:
       - .chameleon_env
     volumes:
@@ -23,7 +28,11 @@ services:
     image: mysql:5.7
     volumes:
       - ./db:/docker-entrypoint-initdb.d
-    restart: always
+    restart: on-failure
+    healthcheck:
+      test: ["CMD", "mysqladmin", "ping", "-h", "localhost"]
+      timeout: 10s
+      retries: 10
     environment:
       MYSQL_ROOT_PASSWORD: ccpass
       MYSQL_DATABASE: chameleon_dev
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -28,3 +28,6 @@ COPY --from=client /project/static/vue /project/static/vue
 COPY --from=client /project/webpack-stats.json /project/webpack-stats.json
 
 EXPOSE 80 443
+
+HEALTHCHECK --interval=1m --timeout=3s \
```
diurnalist, line 32: Port 8890 is the one mapped by docker-compose -- the service itself I think exposes port 8888. Does this still work due to some quirk of how HEALTHCHECK works? Maybe it should be in the docker-compose.yml instead. Which reminds me, does this have any effect on production if it's in the Dockerfile?

*Edit*: Reading more about it, it seems like it should be checking port 8888, but that it's reasonable to keep it in the Dockerfile.

docker-compose.yml
```diff
@@ -23,7 +24,11 @@ services:
     image: mysql:5.7
     volumes:
       - ./db:/docker-entrypoint-initdb.d
-    restart: always
+    restart: on-failure
+    healthcheck:
```
diurnalist, line 28: Nice! Haven't seen a good example of this before.

diurnalist (approved):
Very nice changes :)

docker-compose.yml
msherman64: healthcheck here just overrides whatever is defined in the container.
If you run `docker ps` or `docker-compose ps` you'll see "Up (Healthy)" instead of just up.

It's generally better to have an application specific healthcheck than this sort of trivial check, but this can be a hacky substitute for e.g. properly waiting for the DB to be available.

Dockerfile
msherman64: I wasn't sure here. We `EXPOSE` 80 and 443, but compose is running a custom "command", and mapping 8890->8888?

msherman64: moved to compose file. this really needs to take the entrypoint / cmd into account because of how we use the container.

PR #150, opened 2021-01-21
add exclusions for django migrations

msherman64:
Black and Flake8 config to be compatible with django projects.
Flake8 currently cannot be configured via pyproject.toml,
but many projects already have a tox.ini file.

Line length could be adjusted higher, up to 119 for code, and 79 for
docs as per:
https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,5 @@
 
 [tool.black]
-line-length = 79
+# Set line length explicitly. https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length
+line-length = 88
 exclude = "/*migrations*/"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 
 [tool.black]
-# Set line length explicitly. https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length
+# Set line length explicitly.
+# See https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length for more details
 line-length = 88
 exclude = "/*migrations*/"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,4 @@
-[tool.pylint.messages_control]
-disable = "C0330, C0326"
-
-[tool.pylint.format]
-max-line-length = "88"
 
 [tool.black]
-line-length = 88
+line-length = 79
+exclude = "/*migrations*/"
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [flake8]
 # line length is intentionally set to 80 here because black uses Bugbear
-# See https://github.com/psf/black/blob/master/docs/the_black_code_style.md#line-length for more details
+# See https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length for more details
 max-line-length = 80
 # Enable bugbear opitionated warnings for line length
 select = C,E,F,W,B,B950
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -0,0 +1,6 @@
+[flake8]
+#Black Compat
+max-line-length = 79
+extend-ignore = E203, W503
+#Django Compat
+exclude = .git,*migrations*
```
diurnalist, line 6: I thought general PEP 8 guidelines dictated newlines at the end of a file. This isn't a python file, but maybe good to keep that as a standard. I have my editor set up to auto-add this anyways, so any file I touch will get one added.

diurnalist (approved):
Makes sense! 🚢

tox.ini
msherman64: This is from: [Black Compatible Configs: flake8 ](https://black.readthedocs.io/en/stable/compatible_configs.html#id2)

> When breaking a line, Black will break it before a binary operator. This is compliant with PEP 8, but this behaviour will cause flake8 to raise W503 line break before binary operator warnings.
In some cases, as determined by PEP 8, Black will enforce an equal amount of whitespace around slice operators. Due to this, Flake8 will raise E203 whitespace before ':' warnings.
Since both of these warnings are not PEP 8 compliant, Flake8 should be configured to ignore these warnings via extend-ignore = E203, W503.

msherman64: Oops, misread your comment. I agree, I think I messed it up during git munging.

pyproject.toml
```diff
@@ -1,8 +1,4 @@
-[tool.pylint.messages_control]
-disable = "C0330, C0326"
-
-[tool.pylint.format]
-max-line-length = "88"
 
 [tool.black]
-line-length = 88
+line-length = 79
```
msherman64, line 3: Open question here w.r.t. line length. 

[Black Line Length](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length)
> Black defaults to 88 characters per line, which happens to be 10% over 80.
> If you’re using Flake8, you can bump max-line-length to 88 and forget about it. Alternatively, use Bugbear’s B950 warning instead of E501 and keep the max line length at 80 which you are probably already using.

[flake8 bugbear](https://github.com/PyCQA/flake8-bugbear#opinionated-warnings)
> B950: Line too long. This is a pragmatic equivalent of pycodestyle's E501: it considers "max-line-length" but only triggers when the value has been exceeded by more than 10%. You will no longer be forced to reformat code due to the closing parenthesis being one character too far to satisfy the linter. At the same time, if you do significantly violate the line length, you will receive a message that states what the actual limit is.

[pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)
> Limit all lines to a maximum of 79 characters.
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.
> For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

[Django coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/)
> An exception to PEP 8 is our rules on line lengths. Don’t limit lines of code to 79 characters if it means the code looks significantly uglier or is harder to read. We allow up to 119 characters as this is the width of GitHub code review; anything longer requires horizontal scrolling which makes review more difficult. This check is included when you run flake8. Documentation, comments, and docstrings should be wrapped at 79 characters, even though PEP 8 suggests 72.

diurnalist (approved):
Looks good, I'm going to merge and see how my PR looks now :)

PR #151, opened 2021-01-21
Feature/django rest framework

msherman64:
Replace custom serialization and DB logic with django-rest-framework

Overview: Django Rest Framework provides an extensible method for mapping between Django models and serialized APIs.

We are able to re-use all of the type checking and verification of the models to build standard Create/Read/Update/Delete methods, as well as leverage a robust templating system.

This can replace almost of the custom logic in utils/project_allocation_mapper.py, as well as much of the Vue.js frontend logic.

Edit: Removed extra tasks, for later PR.
Tasks to accomplish this:
- [x] add rest_framework and django-filter as dependencies
- [x] create new logic using rest framework: querysets, serializers, filters, views

Example endpoints:

* `/admin/allocations/allocation-list?format=json&limit=200&status=active`
* `/admin/allocations/project-list?format=json&limit=200&allocations__status=pending`


diurnalist:
Can you give a high-level summary of what we'll be using django-rest-framework for? I'm unfamiliar with it. From perusing the WIP, it looks like it can be a nice improvement and will be generalizable to many things we do.

util/views.py
```diff
@@ -0,0 +1,69 @@
...
+class ListBaseView(ListAPIView):
+    authentication_classes = [SessionAuthentication]
+    permission_classes = [IsAdminUser]
+
+    pagination_class = PaginationBaseView
+    renderer_classes = [
+        BrowsableAPIRenderer,
+        JSONRenderer,
+    ]
+    filter_backends = [DjangoFilterBackend]
```
msherman64, line 35: This is probably the easiest place to start.
Any view inheriting from this will get:

1. session authentication + admin only permission
2. pagination with the defaults we've set above, or in the main config file
3. a browsable API view, that will expose whatever features the class is configured with
4. raw json output

util/views.py
```diff
@@ -0,0 +1,69 @@
...
+    pagination_class = None
+    # Force only JSON renderer for app compatibility
+    renderer_classes = [JSONRenderer, BrowsableAPIRenderer]
+
+    serializer_class = NestedProjectSerializer
+    filterset_class = ProjectFilter
+
+    def get_queryset(self):
+        qs = CCQuerySets()
+        return qs.orderedProjects()
```
msherman64, line 51: Here, the base class is configured to disable pagination, and default to json output if `?format=api` is not specified.
The rest of the behavior is entirely defined by the:

* Serializer class: defines all the logic for how the model / other data structures is serialized and validated.
* filterset class: what dynamic filters are exposed from the URL, and applied to the queryset
* queryset class: the DB level data source.

util/serializers.py
```diff
@@ -0,0 +1,78 @@
...
+            "charge_code",
+            "title",
+            "nickname",
+            "type",
+            "project_type",
+            "field",
+            "description",
+            "pi",
+            "allocations",
+        ]
```
msherman64, line 69: The serializers are the big part of the motivation here. It's most of the hard part of what we've been doing in the "allocation mapper" functions, but formalized, and can automatically infer these properties from the defined models.

This lets us keep all the translation and validation logic in one place.

By default, it will use the primary key to reference foreign keys.

If we define create() or update() methods, these will enable validated write access back to the django DB.

util/serializers.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class NestedProjectSerializer(ProjectSerializer):
+    """Serialize project object, recursively nesting values"""
+
+    project_type = serializers.CharField(source="type.name")
+    field = serializers.CharField(source="field.name")
+    pi = UserSerializer()
+    allocations = NestedAllocationSerializer(many=True)
```
msherman64, line 78: For compatibility with the existing frontend, we can nest the related objects inside the structure instead of just listing a primary key.
We don't need to worry about the querysets and DB optimization, as that logic is self contained.

util/querysets.py
```diff
@@ -0,0 +1,43 @@
...
+    def orderedProjects(self):
+        """Get projects, sorted by latest allocation."""
+        filtered_qs = self.orderedAllocations()
+        project_qs = self.projectsQuerySet()
+        output_qs = (
+            self.prefetchAllocations(filtered_qs, project_qs)
+            .annotate(latest_request=Max("allocations__date_requested"))
+            .order_by("-latest_request")
+        )
+        return output_qs
```
msherman64, line 43: this is not strictly related to the django-rest-framework, but by putting the querysets in one module, we're able hide the complexity of the DB optimizations, and then re-use them as needed.

util/querysets.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def projectsQuerySet(self):
+        """Get all projects from django DB."""
+        logger.debug("Projects QS INIT")
+        output_qs = Project.objects.select_related("type", "pi", "field")
+        return output_qs
+
+    def orderedProjects(self):
+        """Get projects, sorted by latest allocation."""
+        filtered_qs = self.orderedAllocations()
```
msherman64, line 36: This line ensures that the allocations are sorted, before they're nested into the project query.

If we wanted to restrict the listed allocations to only ones matching a query, the filter could be applied here.

util/filters.py
```diff
@@ -0,0 +1,44 @@
...
+    field__name = filters.AllValuesFilter()
+    pi__username = filters.CharFilter()
+    pi__email = filters.CharFilter()
+
+    allocations__status = filters.AllValuesFilter()
+    allocations__su_used = filters.NumericRangeFilter()
+
+    class Meta:
+        model = Project
+        fields = ["charge_code", "title", "nickname", "description"]
```
msherman64, line 44: This module provides a place to define all the valid filter rules that the frontend can ask for. All of the filters are type-checked against the model, and we can define foreign key relations using the standard django syntax.

msherman64:
All the logic in django-filters and filters.py  can be replaced with the package dynamic_rest.

Queries of the form: /api/allocations?include[]=project.*&filter{su_used.gt}=0&sort[]=-date_requested

util/querysets.py
```diff
@@ -0,0 +1,43 @@
...
+from allocations.models import Allocation
+from django.db.models import Max, Prefetch
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class CCQuerySets:
+    def allocationsQuerySet(self):
```
diurnalist, line 11: Style is to do `camel_case`

util/querysets.py
diurnalist: Where exactly is this used? I don't see any references to it in the codebase.

util/querysets.py
```diff
@@ -0,0 +1,43 @@
+from projects.models import Project
+from allocations.models import Allocation
+from django.db.models import Max, Prefetch
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class CCQuerySets:
```
diurnalist, line 10: This file feels like it's automagically doing something just based on its name. Is that true? Are these being loaded/exposed somehow? If so, can we document that this is the case at the top of the file.

util/querysets.py
diurnalist: I think this maybe answers some of the above questions, but it's still not clear where this is being used--nowhere yet?

chameleon/settings.py
```diff
@@ -244,6 +247,33 @@
 
 LOCALE_PATHS = (os.path.join(BASE_DIR, "locale"),)
 
+
+REST_FRAMEWORK = {
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "rest_framework.authentication.SessionAuthentication"
+    ],
+    # if not set explicitly, make admin only
```
diurnalist, line 255: 👍🏻 -- how do authentication credentials work? Is it token-based? "SessionAuthentication" seems to me that it's working based on the Django cookie.

chameleon/settings.py
```diff
@@ -244,6 +247,33 @@
...
+    ],
+    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
+    "PAGE_SIZE": 100,
+}
+
+DYNAMIC_REST = {
+    # DEBUG: enable/disable internal debugging
+    "DEBUG": False,
+    # ENABLE_BROWSABLE_API: enable/disable the browsable API.
+    # It can be useful to disable it in production.
```
diurnalist, line 272: Should we disable it in production as recommended? What about setting these to be derived from the value of `DEBUG`?

chameleon/settings.py
```diff
@@ -292,6 +322,11 @@
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
+    "filters": {
+        "require_debug_true": {
```
diurnalist, line 326: Cool, I didn't know about this mechanism :)

chameleon/urls.py
```diff
@@ -1,100 +1,151 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+router = routers.DynamicRouter()
+router.register(r"api/projects", ProjectViewSet)
+router.register(r"api/users", UserViewSet)
+router.register(r"api/allocations", AllocationViewSet)
+router.register(r"api/types", TypeViewSet)
+router.register(r"api/fields", FieldViewSet)
```
diurnalist, line 34: 👍🏻 -- I agree that's a good place for them.

util/dynamic_drf_api.py
```diff
@@ -0,0 +1,132 @@
...
+
+class AllocationSerializer(DynamicModelSerializer):
+    """Serialize Allocation object."""
+
+    """
+    keys in the array "fields" can be queried in URL by "?key=value"
+    Their format can be overridden by defining it outside the "Meta" class.
+    """
+
+    # if self.instance.status == "active":
```
diurnalist, line 40: Yeah, I was going to ask how this was going to work. Is it possible to make a single batch fetch to the balance service as part of serializing the allocations?

diurnalist (approved):
This looks like a nice change to me! I left some comments. I think the biggest value right now is having pagination just sorted out as a core concern.

BTW, is it easy to call these paged querysets from other views? Or are they mostly intended for XHR / external API endpoints?

util/querysets.py
msherman64: This is documented at: https://docs.djangoproject.com/en/1.11/topics/logging/#naming-loggers

> By convention, the logger name is usually __name__, the name of the python module that contains the logger. This allows you to filter and handle logging calls on a per-module basis. However, if you have some other way of organizing your logging messages, you can provide any dot-separated name to identify your logger

chameleon/settings.py
msherman64: > 👍🏻 -- how do authentication credentials work? Is it token-based? "SessionAuthentication" seems to me that it's working based on the Django cookie.

Docs: https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication

It's directly using django's session authentication backend, but I'm not familiar with the details. It's worth digging into.

util/dynamic_drf_api.py
msherman64: The need to do a bulk fetch seems dependent on triggering the "list view".

One way to do it would be to add to the view, and have it pass the su_usage to the serializer. 
https://www.django-rest-framework.org/api-guide/serializers/#including-extra-context

Alternately, we could have a celery task to just periodically update the db?

util/querysets.py
msherman64: > This looks like a nice change to me! I left some comments. I think the biggest value right now is having pagination just sorted out as a core concern.
> 
> BTW, is it easy to call these paged querysets from other views? Or are they mostly intended for XHR / external API endpoints?

I need to check. The use case appears to be composing views, and rendering based on primary key / hyperlink references between models. I'll add a good example view that exercises the features.

msherman64:
All of the querysets were in use by the "filtersets", but I found the dynamic_rest package that mostly removes the need to pre-define the filtersets and db queries.

It can still be beneficial to have these, for more specific / optimized views.
For example, the current Allocations Approval app requires having allocation objects nested inside the project object json serialization, with a specific sorting order.

Its probably easier to make a new view that points to querysets.orderedprojects than to reconstruct that query in the frontend.

But for a more dynamic / less defined application, you can instead point to the general list, and call something like:

`/api/projects/?include[]=allocations.*&sort[]-allocations.date_requested&filter{allocations.status}=pending`

msherman64:
The actual ordering of allocation w.r.t. projects, and the like, is better done by defining model properties:
[get_latest_by](https://docs.djangoproject.com/en/1.11/ref/models/options/#get-latest-by)
[order_with_respect_to](https://docs.djangoproject.com/en/1.11/ref/models/options/#order-with-respect-to)
[ordering](https://docs.djangoproject.com/en/1.11/ref/models/options/#django.db.models.Options.ordering)

PR #152, opened 2021-01-22
pretty-print sql logs

msherman64:
This PR fixes sql logging in Django.

We needed to pin the mysql client version.
The Django SQL logs then expose the query explicitly.

Adds a Formatter to optionally pretty-print queries.


diurnalist (approved):
This looks like a great improvement! The query logs were really hard to figure out before when they only printed the query arguments.

PR #154, opened 2021-01-25
Render orcid ID as clickable link if available

msherman64:
Try to parse ORCID ID from availble user info.

TODO
This is a stopgap, we should get authenticated user info
on sign in, and attach it to their accounts.

### DIFF
--- a/chameleon/admin.py
+++ b/chameleon/admin.py
@@ -7,6 +7,7 @@ from functools import wraps
 
 from chameleon.models import PIEligibility
 from django.contrib import admin
+from django.contrib.admin.utils import flatten
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from util.keycloak_client import KeycloakClient
@@ -97,7 +98,7 @@ class PIEligibilityAdmin(admin.ModelAdmin):
         The regex below CANNOT validate that an orcid ID exists, or points to the
         correct user.
         """
-        orcid_id_regex = r"\d{4}-\d{4}-\d{4}-(?:\d{3}X|\d{4})"
+        orcid_id_regex = r"\d{4}-\d{4}-\d{4}-(?:\d{3}[xX]|\d{4})"
 
         def get_orcid_ids(input_str, regex):
             """Return all substrings matching regex."""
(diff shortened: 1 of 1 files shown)


msherman64:
Missing the clickable HREF part, due to complex template in /chameleon/templates/admin/index.html

chameleon/admin.py
```diff
@@ -1,52 +1,156 @@
...
+        def is_orcid_id(test_id, regex):
+            """Check if single ID is orcid ID."""
+            match = re.fullmatch(regex, test_id)
+            logger.debug(f"ORCiD Check on {test_id} results in: {match}")
+            return match
+
+        def get_orcid_url(orcid_id):
+            """Return url of form https://orcid.org/0000-0012-2345-6789."""
+            base_url = "https://orcid.org/"
+            if is_orcid_id(orcid_id, orcid_id_regex):
```
diurnalist, line 117: I'm not sure you need this, given this is an internal function that is only called once it's established the IDs are valid ORCiDs.

chameleon/admin.py
```diff
@@ -1,52 +1,156 @@
...
+                logger.debug(f"converting {orcid_id} to {result}")
+                return result
+            else:
+                logger.debug(f"could not map {orcid_id} to URL")
+                return None
+
+        # Get all matching substrings
+        checked_ids = []
+        for user_id in [obj.requestor.username, obj.requestor.email]:
+            orcid_ids = get_orcid_ids(user_id, orcid_id_regex)
```
diurnalist, line 128: This could be simplified using a few things in Python, namely itertools.chain and list comprehension.

```python
from itertools import chain
search_keys = [obj.requestor.username, obj.requestor.email]
orcid_ids = [oid for oid in chain(*[get_orcid_ids(search, orcid_id_regex) for search in search_keys) if oid]
```

The `if oid` is only necessary if the list can have None or falsy values in it.

diurnalist (approved):
LGTM, but what is the issue with rendering the link as HTML, precisely?

chameleon/admin.py
msherman64: This was being a bit defensive, for the case where we're getting possible IDs from multiple sources, e.g. parsing username, email, general user input, metadata from keycloak, etc.

chameleon/admin.py
msherman64: used flatten() to avoid needing to handle [ [], [], ] that regex would spit out occassionally.

msherman64:
> LGTM, but what is the issue with rendering the link as HTML, precisely?

I'm not sure 

> LGTM, but what is the issue with rendering the link as HTML, precisely?

In the actual detail field, e.g. `http://localhost:8890/admin/chameleon/pieligibility/795/change/`

I cannot seem to find where the "pieligibility_form" is defined.

PR #155, opened 2021-01-26
fix broken link to faq from project creation page

msherman64:


msherman64:
Should the standard be:
All documentation links are to the absolute URL, not relative path?

msherman64:
Resolve [RT#67501](https://consult.tacc.utexas.edu/Ticket/Display.html?id=67501)

PR #156, opened 2021-01-26
Use same .env file for docker-compose and makefile.

msherman64:
There's a bit of noise in the PR due to the auto-formatter, I can clean that up into two commits if needed.

The purpose of this PR is to:
1. Clean up the logging configuration for easier debugging.
2. Put configuration paramaters that are common to both the makefile, docker-compose file, and running container, into the .chameleon_env file
3. rename .chameleon_env to .env, so that docker-compose will automatically use it without specifying --env-file
4. add a valid .sample.env to the repo, so that local development has fewer manual steps
5. add a target to the makefile to copy .sample.env -> .env if missing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,19 @@
-#Set language versions
-PY_IMG_TAG=3.7.9-stretch
-NODE_VER=lts
-
-DOCKER_REGISTRY ?= docker.chameleoncloud.org
-DOCKER_TAG ?= $(shell git rev-parse --short HEAD)
-DOCKER_IMAGE := $(DOCKER_REGISTRY)/portal:$(DOCKER_TAG)
-DOCKER_IMAGE_LATEST := $(DOCKER_REGISTRY)/portal:latest
-
-PORTAL_MANAGE_CMD := docker-compose exec portal python manage.py
-#if APP is unset, then the variable equals the empty string.
-ifeq ($(APP),)
-MIGRATIONS_CMD := $(PORTAL_MANAGE_CMD) makemigrations
-else
-MIGRATIONS_CMD := $(PORTAL_MANAGE_CMD) makemigrations "$(APP)"
+# Set make variables from .env file
+ifneq (,$(wildcard ./.env))
+    include .env
+    export
+	ENV_FILE_PARAM = --env-file .env
 endif
 
+DOCKER_TAG ?= $(shell git rev-parse --short HEAD)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,22 +1,19 @@
...
-#if APP is unset, then the variable equals the empty string.
-ifeq ($(APP),)
-MIGRATIONS_CMD := $(PORTAL_MANAGE_CMD) makemigrations
-else
-MIGRATIONS_CMD := $(PORTAL_MANAGE_CMD) makemigrations "$(APP)"
-endif
+DOCKER_IMAGE ?= $(DOCKER_REGISTRY)/portal:$(DOCKER_TAG)
+DOCKER_IMAGE_LATEST ?= $(DOCKER_REGISTRY)/portal:latest
+
+.env:
```
diurnalist, line 12: I was going to suggest this use `.env.sample` as a dependency, then realized that would mean .env would get overridden when the sample changed. Yes, good to keep without dependencies then :)

chameleon/settings.py
```diff
@@ -1,4 +1,6 @@
 """
+Django settings file. This gets most values by loading a .env file.
```
diurnalist, line 2: Don't know that we necessarily have to say the env variables are coming from a dotenv file, but as documentation this is fine!

chameleon/settings.py
```diff
@@ -319,6 +319,15 @@
 # Logger config
 #
 #####
+
+# DEBUG = True if DJANGO_ENV_DEBUG==DEBUG
```
diurnalist, line 323: Did you mean to comment this out still?

docker-compose.yml
```diff
@@ -5,18 +5,19 @@ services:
...
     volumes:
       - .:/project
       - ./media:/media
       - static:/static
     ports:
-      - 127.0.0.1:8890:8888
-    command: python3 manage.py runserver 0.0.0.0:8888
+      - 127.0.0.1:8890:${PORTAL_PORT}
+    entrypoint: ["python3", "manage.py"]
+    command: ["runserver", "0.0.0.0:${PORTAL_PORT}"]
```
diurnalist, line 20: Nice 👍🏻

diurnalist (approved):
Great change! This is a nice improvement, much cleaner now. It will break any local dev environments still using .chameleon_env as the primary env. I think that's OK... just letting @zhenz know as well :)

chameleon/settings.py
msherman64: That was informational, not code. Replaced with links to the django debug setting docs.

PR #161, opened 2021-02-11
Use Github actions to test portal build

msherman64:
This was a naive look at building the portal container with GH actions.

The v2 docker actions now support caching using buildkit, allowing us to use a single multi-stage build
without speed penalties.

TODO:
Figure out how to import a "sanitized" db, and run migration tests.

TODO: Publish containers somewhere.

### DIFF
--- a/.github/workflows/build-portal.yml
+++ b/.github/workflows/build-portal.yml
@@ -9,9 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


diurnalist:
> Figure out how to import a "sanitized" db, and run migration tests.
Do we need a DB? The migrations should at least test that the schema is migrated. Not having a DB means that live data migrations won't be tested. But I'm not sure of a scalable way to do this. Maybe we can just have migration tests for domain objects under our control (like Trovi's models.)

PR #166, opened 2021-03-02
Feature/blog slider

msherman64:
Two changes here.
The first is just a cleanup of the env.sample file, maybe it should be a different PR.

Second, adding the module `aldryn-bootstrap3` to use the carousel example from http://docs.django-cms.org/en/user-tutorial/user/tutorial/100-key-plugins.html

There are better slider plugins, but that should probably wait until we have the cycles to really work on portal.
We'd want to, at a minimum, look at upgrading to django2.2 and bootstrap4 for broader support there.


Makefile
```diff
@@ -1,7 +1,7 @@
 # Set make variables from .env file
 ifneq (,$(wildcard ./.env))
-    include .env
-    export
+	include .env
```
diurnalist, line 3: Huh, I'm surprised this was not caught before. Tabs vs. spaces?

diurnalist (approved):
Looks good to me! And, agree that we should earmark some time for the Django 2.2 (or 3?) upgrade.

Makefile
msherman64: Possibly. I changed my editor a bit.

PR #167, opened 2021-03-08
move carousel to js file

msherman64:
use glide.js for slider.

use webpack to load css files and js for glide.js.


msherman64 (commented):
w.r.t. the extra files, I was following the instructions here:
https://djangocms-blog.readthedocs.io/en/latest/features/templates.html#plugin-templates

> When using this feature you must provide all the templates for the available plugins.

diurnalist (approved):
Re: the instructions, I am not sure they apply in this case. Definitely all the templates need to be there for this plugin to be generally workable. But it appears it doesn't have to have all of them to work. I think this whole thing is a bit of a holdover. There doesn't seem to be much benefit of having this be CMS plugin (or the front page to be a CMS page.) I would rather we just render the home page like we do other pages; that way we don't have to fool around with this stuff.

PR #169, opened 2021-03-09
Fixup RT ticket views

diurnalist:
This is a general refactor of the RT app, but just the view layer.
This moves most of the ticket handling logic to a core function, which
is called for both the logged in and guest case. Also fixes a bug where
mimetype was set to a tuple and not a string, which apparently caused
RT to throw its hands in the air and not process the file properly.

Added some better/more consistent error messaging, and ensured that
Python warnings were making it to the logging infra--the RT client uses
this mechanism to emit errors, so it's important we are seeing them.

### DIFF
--- a/djangoRT/views.py
+++ b/djangoRT/views.py
@@ -6,199 +6,235 @@ from django.contrib.auth.decorators import login_required
 from django.contrib import messages
 import logging
 import mimetypes
-from .models import TicketCategories
-from wsgiref.util import FileWrapper
 
-from keystoneclient import client as ks_client
-from keystoneauth1.identity import v3
-from keystoneauth1 import adapter, session
-import json
 from novaclient import client as nova_client
 from blazarclient import client as blazar_client
 from glanceclient import Client as glance_client
-from django.conf import settings
-from datetime import datetime
 from dateutil import parser
-import sys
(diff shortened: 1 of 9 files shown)


djangoRT/views.py
```diff
@@ -6,200 +6,236 @@
...
-            if ticket_id > -1:
-                if 'attachment' in request.FILES:
-                    rt.replyToTicket(ticket_id, files=([request.FILES['attachment'].name, request.FILES['attachment'], mimetypes.guess_type(request.FILES['attachment'].name)],))
-                messages.add_message(request, messages.SUCCESS, 'Ticket #%s has been successfully created. We will respond to your request as soon as possible.' % ticket_id)
-                form = forms.TicketGuestForm()
-                return render(request, 'djangoRT/ticketCreateGuest.html', { 'form': form })
-            else:
-                # make this cleaner probably
-                messages.error('An unexpected error occurred while creating your ticket. Please try again.')
-                data['first_name'] = form.cleaned_data['first_name']
```
diurnalist, line 136: This kind of thing is not necessary. The form will default to the values it found on POST and if it's not valid, it will re-render those values back in to the form with error messages. This is what Django is designed to do.

PR #171, opened 2021-03-23
set default su and start/end dates

zhenz:
Default SU, start and end dates when approving allocation


zhenz:
@diurnalist 

Here's the lines of code I added
`        var g = angular.copy(b)
          , i = f.open({
            templateUrl: "/admin/allocations/template/approval.html/",
            controller: "modalController",
            resolve: {
                data: function() {
                    if (g.computeAllocated == null) {
                        g.computeAllocated = g.computeRequested;
                    }
                    var now = new Date();
                    if (g.start == null) {
                        g.start = now.toISOString();
                    }
                    if (g.end == null) {
                        g.end = new Date(now.setMonth(now.getMonth() + 6)).toISOString();
                    }
                    return {
                        allocation: g,
                        type: "approve"
                    }
                }
            }
        });`

diurnalist (approved):
That code looks good to me, there may be some way of doing this w/ just Angular templating but I am not sure how. If this looks good on your local dev, 🚢 !

PR #172, opened 2021-03-25
allow amending allocation reviews

zhenz:


zhenz:
@diurnalist I only allowed switching decision from approved to reject, and reject to approve.

diurnalist (approved):
How does the decision summary get updated? Do you update the summary and then click one of these buttons, and it sends the updated text in a notification?

zhenz:
> How does the decision summary get updated? Do you update the summary and then click one of these buttons, and it sends the updated text in a notification?

You click the button, and there will be a modal (popup window) for you to update the decision summary.

PR #174, opened 2021-03-30
allow user select project type

zhenz:


zhenz:
@diurnalist changes are related to categorizing projects into cs research, education, and innovative applications. I reused the "type" field for this. We need a deployment plan for this for all the old projects. Currently, what I'm thinking is to 
 1. update projects_type table manually to the three new types
 2. assign projects which have "course" in the description as education type
 3. assign projects which have "COVID" in the description as innovative applications
 4. assign the rest of the projects as cs research
There will be miscategorized projects. We might need to manually correct those.

diurnalist:
Are those types carried over from TAS or did you discuss them with outreach? In principle I think it's fine, and it's the right thing to re-use! We can adjust types later. I think COVID should potentially be its own type actually, it's a unique enough event that it will be useful to see a collection of projects w/ just that type.

projects/forms.py
```diff
@@ -86,6 +86,12 @@ class ProjectCreateForm(forms.Form):
         initial="45",
         help_text="Please indicate a primary field of science for this research.",
     )
+    typeId = forms.ChoiceField(
+        label="Type",
+        choices=(),
+        initial="1",
```
diurnalist, line 92: What does "1" refer to? What happens if it is simply None, or left out?

diurnalist (approved):

projects/forms.py
```diff
@@ -86,6 +86,12 @@ class ProjectCreateForm(forms.Form):
         initial="45",
         help_text="Please indicate a primary field of science for this research.",
     )
+    typeId = forms.ChoiceField(
+        label="Type",
+        choices=(),
+        initial="",
+        help_text="Please indicate a type of this project.",
```
diurnalist, line 93:
```suggestion
        help_text="Please indicate a project type.",
```

projects/templates/projects/view_project.html
```diff
@@ -46,7 +46,21 @@ <h2>
       </tr>
       <tr>
         <th>Type</th>
-        <td>{{ project.type }}</td>
+        <td>
+          {% if is_admin %}
+            <span id='project_type'>{{ project_type|default_if_none:'' }}</span><span id='icon_show_type_edit' title="Edit" style="padding-left: 5px;" class="glyphicon glyphicon-edit"></span>
+            <span id='project_type_edit_form' class='hidden'>
```
diurnalist, line 52: You could probably set the ID of the `<form>` to `project_type_edit_form` and put the hidden class on that as well, then you don't need this `<span>`, I think?

projects/views.py
```diff
@@ -484,6 +494,36 @@ def edit_nickname(request, project_id):
     return form
 
 
+
+@require_POST
+def edit_type(request, project_id):
+    form_args = {"request": request}
+    if not request.user.is_superuser:
```
diurnalist, line 501: Very nice, I immediately started looking for this when i saw the `{% is_admin %}` in the template :) -- good that we are checking on both the client and the server.

projects/views.py
```diff
@@ -484,6 +494,36 @@ def edit_nickname(request, project_id):
...
+    form = EditTypeForm(request.POST, **form_args)
+    if form.is_valid(request):
+        # try to update type
+        try:
+            project_type_id = int(form.cleaned_data["typeId"])
+            ProjectAllocationMapper.update_project_type(project_id, project_type_id)
+            form = EditTypeForm(**form_args)
+            messages.success(
+                request,
+                "Update successful! Click <a href='{}'>here</a> to reload".format(
```
diurnalist, line 514: Why is this reload needed? Does the page not display the type entered?

projects/views.py
```diff
@@ -484,6 +494,36 @@ def edit_nickname(request, project_id):
...
+                    request.path
+                ),
+            )
+        except Exception:
+            messages.error(request, "Failed to update project type")
```
diurnalist, line 519: I would also log this exception, otherwise it will be hard to know what happened. You can use `LOG.exception` for that.

diurnalist (approved):
Looks good to me! A few nits and suggestions, but 👍🏻

projects/views.py
zhenz: no. it's not. same for nickname.

projects/templates/projects/view_project.html
zhenz: that's for hiding the form and two buttons - "update" and "cancel" buttons.

PR #185, opened 2021-04-15
add a new allocation status

zhenz:
A new allocation status -- "waiting" is added. Allocation requests
with this status are waiting for response from the project PI. In
addition, a RT ticket will be created when switching from "pending"
to "waiting" to contact the PI for more information.


allocations/migrations/0005_auto_20210414_1610.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
diurnalist, line 1: Can you please rename the part after `0005_` to something that is more descriptive? The auto generated names make it hard to tell what exactly the migration is doing.

allocations/static/allocations/js/main.js
```diff
@@ -1 +1,750 @@
...
+            "error" === a.data.status ? (d.addMessage(c, e, "danger"),
+            null) : (d.addMessage(c, "This allocation request is approved successfully.", "success"),
+            a.data.result)
+        }, function() {
+            d.removeLoading(c),
+            d.addMessage(c, e, "danger")
+        })
+    }
+    ,
+    e.waitingAllocation = function(b) {
```
diurnalist, line 291: This is the main change to this file, right? Man, this is so painful. I declare a moratorium on more JS updates until we rewrite this in something that has a supported build chain!

diurnalist: Can you also include this change in to whatever source file this generated file came from? That way we don't accidentally use it when porting this functionality later.

diurnalist (approved):
This looks good to me! Nice work. Please do add your JS changes to the source file used to generate that final file in /static, which should be elsewhere in Portal.

PR #188, opened 2021-04-21
create ticket and send message

zhenz:
The implementation before this commit will create a RT ticket but not
send the message to the PI. Need to create a ticket, and then reply.

### DIFF
--- a/util/project_allocation_mapper.py
+++ b/util/project_allocation_mapper.py
@@ -114,11 +114,15 @@ class ProjectAllocationMapper:
         )
         ticket = rtModels.Ticket(
             subject=subject,
-            problem_description=problem_description,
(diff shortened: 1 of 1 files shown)


zhenz:
> OK, this isn't as elegant as we'd like, but it gets the job done. So there was no way to have it send to the person another way? What if the requestor/owner were swapped (does it make sense)?

good idea. I'll try.

zhenz:
> > OK, this isn't as elegant as we'd like, but it gets the job done. So there was no way to have it send to the person another way? What if the requestor/owner were swapped (does it make sense)?
> 
> good idea. I'll try.

@diurnalist swapping requestor and owner doesn't work. the owner has to have a TACC account, and can't be an email address. I tried different things with the python rt library, but this is the only way I could think of to solve the problem.

PR #189, opened 2021-04-22
refactor dockerfile and update dependencies

msherman64:
With docker-buildkit, multistage builds can now execute in parallel.
This refactors the separate client dockerfile into a single multistage build for easier CI integration.

In addition, some dependencies are pulled into package.json and pyproject.toml rather than being hardcoded in the dockerfile.


diurnalist (approved):
Look like good changes to me! I had the client build out in a separate folder largely for docker context cache reasons... fewer files that could be affected == less likely the need is to rebuild it.

PR #192, opened 2021-04-26
Invite users to projects via email

msherman64:
Currently before being added to a project, a user must already exist in the system. This can make it a bit annoying for teachers or PIs who want to enroll several users new to Chameleon, as they must wait until the user logs in for the first time.

If it was possible to send invitations to a user's email, and have the invitation add them to the project automatically after they logged in the first time, onboarding of new projects could be significantly improved.
Approach

Create a new DB model to hold invitations. This model should keep track of the user that created the invite (the PI / delegate), the date it was issued, a project ID, whether it has been "accepted", and when it was accepted, and by what user. It should also have an expiration date, default... 30 days? And, a base64-encoded 20+byte random string, which will be used for the invite code. (Read this for why this is better than a UUID in this case.)

Add an admin view to manage invitations as a staff member. This should be a straightforward admin form built around the DB model.

Create a web endpoint to accept an invitation. The endpoint should have the invite code in the path or request arguments. It should require a login (which will require the user sign up as well, an added bonus.) The endpoint should add the user to the invite project's Keycloak group and then take them back to their project dashboard with a message confirming they've been added. Set the status of the invite to accepted, and set the date it was accepted, and the user accepting the invite. For simplicity, it should be possible to accept multiple invites for the same project w/o error.

Add a form option for PIs to invite a user by typing an email address. The form should accept an email address and send an email to the target user synchronously, and return an error if the mail could not be sent or the email address didn't validate. Do not allow sending another invite to the same email address.

Display pending invitations in project member list. Show pending invitations somehow. Allow canceling an invite; this should delete the invite from the DB. Optionally put a "resend" link that triggers re-sending of the existing email invite.

Create a periodic job to delete expired invitations that have not already been accepted. Should be self-explanatory.

- [x] Create a new DB Model to hold Invitations
- [x] Add an admin view to manage invitations as a staff member
- [x] Create a web endpoint to accept an invitation
- [x] add a form option for PIs to invite a user by typing an email address
- [x] Display pending invitations in project member list
- [x] create a periodic job to delete expired invitations that have not already been accepted


projects/models.py
```diff
@@ -78,6 +92,97 @@ class ProjectExtras(models.Model):
...
+
+    def _generate_expiration():
+        now = timezone.now()
+        duration = timezone.timedelta(days=Invitation.default_days_until_expiration())
+        return now + duration
+
+    # This information is needed on creation
+    project = models.ForeignKey(Project, on_delete=models.CASCADE)
+    user_issued = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.DO_NOTHING, editable=True
```
diurnalist, line 133: What does `editable` mean here? Is that something for the admin UI? If so, why is it editable?

projects/models.py
```diff
@@ -78,6 +92,97 @@ class ProjectExtras(models.Model):
...
+        return not self._is_accepted() and not self._is_expired()
+
+    def _is_accepted(self):
+        return self.status == "ACCEPTED"
+
+    def _is_expired(self):
+        return self.date_expires < timezone.now()
+
+    def save(self, *args, **kwargs) -> None:
+        return super().save(*args, **kwargs)
```
diurnalist, line 183: This is redundant

projects/models.py
```diff
@@ -78,6 +92,97 @@ class ProjectExtras(models.Model):
...
+
+
+class InvitationManager(models.Manager):
+    pass
+
+
+class Invitation(models.Model):
+    """Model to hold invitations of users to projects."""
+
+    STATUS = [("ISSUED", "Issued"), ("ACCEPTED", "Accepted")]
```
diurnalist, line 106: Can you put the values ("ISSUED", "ACCEPTED") as constants, which you can then refer to in other places? This will make it harder to screw up referencing them in some other class/elsewhere. For example, in the celery task.

projects/templates/projects/view_project.html
```diff
@@ -149,6 +149,36 @@ <h4>Current Users</h4>
...
+        <button type="submit" class="btn btn-xs btn-link" name="resend_invite"><i class="fa fa-envelope text-info"></i><span class="sr-only">Resend invitation</span></button>
+      </form>
+      <form role="form" method="post" style="display:inline">
+        {% csrf_token %}
+        <input type="hidden" name="email_code" value="{{ i.email_code }}">
+        <button type="submit" class="btn btn-xs btn-link" name="del_invite"><i class="fa fa-minus-square text-danger"></i><span class="sr-only">Remove invitation</span></button>
+      </form>
+    </li>
+    {% endfor %}
+  </ul>
```
diurnalist, line 179: Can you attach a screenshot of what this looks like / where it is relative to other UI? I imagined it would be around the project membership section.

projects/templates/projects/view_project.html
```diff
@@ -149,6 +149,36 @@ <h4>Current Users</h4>
   {% endfor %}
 </ul>
 
+{% if is_pi %}
+  <h4>Pending Invitations</h4>
+
+  <form class="form-inline" role="form" method="post">
+    {% csrf_token %}
+    {% bootstrap_form invite_form %}
```
diurnalist, line 157: I would think that `bootstrap_form` already includes both `<form>` and the csrf token. Maybe it doesn't. Check the output HTML to see what it's generating. I think there is also a way to pass the form class in on `bootstrap_form`. 

**Edit**: I looked at the docs https://github.com/zostera/django-bootstrap3 and was surprised that this isn't the case. I do think you can use that "layout" argument to make it inline though, perhaps.

projects/templates/projects/view_project.html
```diff
@@ -149,6 +149,36 @@ <h4>Current Users</h4>
...
+    <div class="form-group">
+      <button type="submit" class="btn btn-default" name="invite_user">Invite user</button>
+    </div>
+  </form>
+
+  <ul class="list-unstyled">
+    {% for i in invitations %}
+    <li>
+      {{i.email_address}}
```
diurnalist, line 166: style nit: always include spaces around curlys

```suggestion
      {{ i.email_address }}
```

projects/templates/projects/view_project.html
```diff
@@ -149,6 +149,36 @@ <h4>Current Users</h4>
...
+      <form role="form" method="post" style="display:inline">
+        {% csrf_token %}
+        <input type="hidden" name="email_code" value="{{ i.email_code }}">
```
diurnalist, line 169: If you're re-sending, it is safer to just include the ID of the invitation here instead of the code. The server can retrieve the code and re-send. No need to expose it in the HTML here, even in a hidden elem.

projects/templates/projects/view_project.html
```diff
@@ -149,6 +149,36 @@ <h4>Current Users</h4>
...
+        <button type="submit" class="btn btn-xs btn-link" name="resend_invite"><i class="fa fa-envelope text-info"></i><span class="sr-only">Resend invitation</span></button>
+      </form>
+      <form role="form" method="post" style="display:inline">
+        {% csrf_token %}
+        <input type="hidden" name="email_code" value="{{ i.email_code }}">
```
diurnalist, line 174: Same point here

projects/views.py
```diff
@@ -1,43 +1,45 @@
-from django.shortcuts import render
-from django.contrib.auth.decorators import login_required
+import json
+import logging
+import re
+import sys
+import uuid
```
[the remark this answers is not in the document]
diurnalist: Can you fixup these 'imported but unused' issues by removing the import?

projects/views.py
```diff
@@ -81,6 +85,23 @@ def user_projects(request):
     return render(request, "projects/user_projects.html", context)
 
 
+@login_required
+def accept_invite(request, invite_code):
+    mapper = ProjectAllocationMapper(request)
+    try:
+        accepted, result = mapper.accept_invite(request.user, invite_code)
```
diurnalist, line 92: Some context: this `ProjectAllocationMapper` was more or less an implementation of the [Strangler pattern](https://martinfowler.com/bliki/StranglerFigApplication.html). We introduced it because we were in a situation where we had to simultaneously support old legacy users and new federated users, but we had tons of code doing common operations against each type of user, and those operations had very different behavior depending on this.

We have since migrated over finally, so we should start trying to tear down some of the complexity in this mapper. For this reason I don't think we should be adding more logic here.
I think it's arguably simpler to just do:

```python
invite = Invitation.get(invite_code=invite_code)  # or similar... see other comments about using ID instead
accepted = invite.accept(request.user)
if accepted:
    # message...
    return HttpResponseRedirect(reverse("projects:view_project", args=[invite.project])
else:
    # message...
```

projects/views.py
```diff
@@ -148,10 +172,40 @@ def view_project(request, project_id):
                     "An unexpected error occurred while attempting "
                     "to remove this user. Please try again",
                 )
+        elif "del_invite" in request.POST:
+            try:
+                email_code = request.POST["email_code"]
+                mapper.remove_invitation(project_id, email_code)
```
diurnalist, line 178: Similar to above, just use the Invitation class to delete the entry.

projects/views.py
```diff
@@ -148,10 +172,40 @@ def view_project(request, project_id):
...
+                logger.exception("Failed to delete invitation")
+                messages.error(
+                    request,
+                    "An unexpected error occurred while attempting "
+                    "to remove this invitation. Please try again",
+                )
+        elif "resend_invite" in request.POST:
+            try:
+                email_code = request.POST["email_code"]
+                mapper.resend_invitation(project_id, email_code, request.user, request.get_host())
```
diurnalist, line 192: This might need a helper function, but let's not put it in the mapper. It can probably just be a helper function in this file.

projects/views.py
```diff
@@ -188,6 +242,16 @@ def view_project(request, project_id):
             logger.info("user: " + u.username + " not found")
         user_mashup.append(user)
 
+    invitations = mapper.get_project_invitations(project_id)
```
diurnalist, line 245: Similarly, use Invitation to look up invites based on project.

projects/views.py
```diff
@@ -464,6 +530,36 @@ def edit_project(request):
     return render(request, "projects/edit_project.html", context)
 
 
+@require_POST
+def invite_user(request, project_id):
+    mapper = ProjectAllocationMapper(request)
+    project = mapper.get_project(project_id)
+    if not project_pi_or_admin_or_superuser(request.user, project):
```
diurnalist, line 537: 👍🏻 good job checking this on both client and server sides.

projects/views.py
```diff
@@ -464,6 +530,36 @@ def edit_project(request):
...
+        messages.error(request, "Only the project PI can invite users.")
+        return InviteUserEmailForm()
+
+    form = InviteUserEmailForm(request.POST)
+    if form.is_valid(request):
+        try:
+            email_address = form.cleaned_data["user_email"]
+            if mapper.email_exists_on_project(project_id, email_address):
```
diurnalist, line 545: This function is harder to relocate from the mapper. Maybe in a utils.py? Could be used in both views/tasks, is why I say that.

projects/views.py
```diff
@@ -464,6 +530,36 @@ def edit_project(request):
...
+                messages.error(request, "That email is tied to a user already"
+                                        "on the project!")
+            else:
+                mapper.add_project_invitation(project_id, email_address, request.user, request.get_host())
```
diurnalist, line 549: Could be function in this views.py file.

util/project_allocation_mapper.py
```diff
@@ -341,6 +343,88 @@ def get_project_members(self, tas_project):
                 users.append(user)
         return users
 
+    def email_exists_on_project(self, project_id, email_address):
```
diurnalist, line 346: Hmm, OK, I see that this one is harder to extract b/c it depends on other mapper functions. We can leave this in here and address it later as part of a larger refactor.

projects/views.py
diurnalist: Edit: I saw that accept_invite is doing more stuff, notably actually adding the project membership. What if you _only_ used the mapper to add the project membership, but everything else was done here, or in a helper function in this file? That would keep the mapper's use minimized and make it easier to refactor later.

diurnalist (changes requested):
This is almost there, the main thing I think we should do is keep changes to the ProjectAllocationMapper to a minimum. We want to refactor that crufty thing at some point. I also left some comments about not exposing the invite code to the HTML. Just best not to disclose something sensitive like that to the client.

projects/models.py
Mark-Powers: `editable` is for the admin UI, and so it is not needed here.

projects/templates/projects/view_project.html
Mark-Powers: image
Let me know what you think would be best. One thing I just noticed is the input is below the title here, but the "Current Users" title and the input to add new users are opposite of this. We could include more space between the Users and Invites sections.

Mark-Powers: Another thing I notice here is clearing the input after submitting.

projects/templates/projects/view_project.html
Mark-Powers: I tried `layout='inline'` in the bootstrap_form, and seems to be doing something different from `class="form-inline" ` if that is what your edit refers to. I'll leave it as is now, since it keeps it as the "Add User" form looks.

projects/templates/projects/view_project.html
Mark-Powers: Right now, clicking the resend icon generates a new invite object, with a new code, while expiring the previous invitation. I can still update this so that it's the invite ID here, but just want to clarify if this behavior is what we want.

diurnalist (approved):
Looking great! I know we talked about a UI change, but I'm +1 on this now.

🥇

projects/views.py
```diff
@@ -135,15 +134,43 @@ def view_project(request, project_id):
...
                         )
                         form = ProjectAddUserForm()
-                except Exception as e:
+                except User.DoesNotExist:
+                    # Try sending an invite
+                    email_address = form.cleaned_data["user_ref"]
+                    try:
+                        validate_email(email_address)
+                        if email_exists_on_project(project, email_address):
+                            messages.error(request, "That email is tied to a "
```
diurnalist, line 143: I wonder in what cases this is possible? So it would have to be a user that cannot be found in Portal by that email, yet the Keycloak IdP has that user mail on the project. Technically possible, I wonder if we have any cases in practice. We should probably keep the check, just thinking aloud.

diurnalist (approved):
Nice! I think that is a cleaner solution from the UX side. Looks great, let's get it out!

PR #196, opened 2021-05-04
Fix allowed_hosts and CSP=self for recaptcha

msherman64:
Adding recaptcha involved defining CSP_FRAME_SRC, but it was missing "self".
In addition, default keys have been defined for recaptcha. If used, recaptcha will always pass, and display a graphical warning.

In addition, the changes to ALLOWED_HOSTS from the recent domain change no longer allowed setting it from the .env file.
This fixes a typo, and allows setting via .env file.

### DIFF
--- a/chameleon/settings.py
+++ b/chameleon/settings.py
@@ -67,15 +67,17 @@ ZENODO_DEFAULT_ACCESS_TOKEN = os.getenv("ZENODO_DEFAULT_ACCESS_TOKEN")
 
 # TEMPLATE_DEBUG = DEBUG
 
-ALLOWED_HOSTS = [
+_default_allowed_hosts = [
     "127.0.0.1",
     "localhost",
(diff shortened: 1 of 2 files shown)


chameleon/settings.py
```diff
@@ -67,16 +67,18 @@
...
 # TEMPLATE_DEBUG = DEBUG
 
-ALLOWED_HOSTS = [
+_default_allowed_hosts = [
     "127.0.0.1",
     "localhost",
     "chameleon.local",
     "chameleoncloud.org",
     "www.chameleoncloud.org",
-    "dev.chameleon.org",
```
diurnalist, line 76: d'oh!! 🤦🏻

diurnalist (approved):
In what circumstances do we currently set ALLOWED_HOSTS from an env file?

LGTM 👍🏻 👍🏻

PR #209, opened 2021-06-10
Bugfix/deploy collectstatic

msherman64:
fix installation of yuglify, and add tests for migrate and collectstatic.


diurnalist:
cool stuff, I like the sanity checks :)

PR #211, opened 2021-06-24
Add Openstack information in RT ticket description

Mark-Powers:
Project, lease, and instance information is added to the end of
a ticket submitted by a logged in user. If a user is anonymous,
no information is displayed.

The version of python-blazarclient is switched to use Chameleon's
version of this library, so that `list_allocations` can be used.

Also, python-ironicclient is added as a new dependency to get the
provision state and instance uuid of a reserved node. This should
make it easier to find the issue when a node in a user's lease
won't launch.

### DIFF
--- a/djangoRT/templates/djangoRT/project_details.txt
+++ b/djangoRT/templates/djangoRT/project_details.txt
@@ -1,23 +1,57 @@
-
-
 ________________________________ Project Support Details ________________________________
 
 {% for region in regions %}
 Region: {{region.name}}
     {% for project in region.projects %}
-    Project: {{project.name}} ID: {{project.id}}
+    Project: {{project.charge_code}}
+        ID: {{project.id}}
         {% if project.leases %}
         Leases:{% for lease in project.leases %}
             {{lease.name}} - {{lease.status}}
                 Date Range: {{lease.start_date}} - {{lease.end_date}}
-                ID: {{lease.id}}{% endfor %}
+                ID: {{lease.id}}
+                Hosts:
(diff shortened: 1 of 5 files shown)


djangoRT/templates/djangoRT/project_details.txt
```diff
@@ -1,23 +1,57 @@
...
+                    {% endfor %}
+                {% if lease.networks %}
+                Networks:{% for network in lease.networks %}
+                    id: {{ network.id }}
+                        physical_network: {{ network.physical_network }}
+                        segment_id: {{ network.segment_id }}{% endfor %} {% endif %}
+            {% endfor %}
+        {% else %}
+        {% if project.lease_error %}
+        Error getting leases for project
```
diurnalist, line 30: 👍🏻

djangoRT/views.py
```diff
@@ -276,73 +292,157 @@ def ticketattachment(request, ticket_id, attachment_id):
...
-        "1", service_type="reservation", interface="publicURL", session=psess
+        "1", service_type="reservation", interface="publicURL", session=sess
     )
+    ironic = ironic_client.Client("1", session=sess)
     leases = blazar.lease.list()
     for lease in leases:
+        if lease.get("user_id") != user_id or lease.get("project_id") != project_id:
+            continue
+
+        # Ignore if more than 3 days out of date
```
diurnalist, line 380: Very nice!

diurnalist (approved):
Very good, I wouldn't change a thing :)

PR #212, opened 2021-06-25
Update allocation expiration date is set if multiple allocations exist

Mark-Powers:
If multiple allocations exist, previously the one with the earlier
expiration date is made inactive. This updates the allocation
expiration date along with this inactive update.


diurnalist (approved):
Nice and easy 👍🏻

PR #215, opened 2021-07-02
Fix next param getting lost on login redirect

Mark-Powers:
When redirected to login for a page that requires it, the next URL
parameter is used to remember the place. However, this was lost
when issuing a redirect to 'oidc_authentication_init'. This patch
fixes this issue by manually adding it to the oidc path when
generating the redirect.


diurnalist (approved):
Looks good, next time we are in here we should completely remove this custom_login thing though, and just configure oidc_authentication_init as the default login handler.

PR #217, opened 2021-07-07
Warn users when allocation is expiring soon

super-cooper:
Implements a new Celery task in the allocations app which warns PIs when their project's allocation is within 1 month of expiring. The email will only be sent once, meaning that the vast majority of users going forward will receive the email exactly one month out from their allocation's expiration date.

Regardless of when the mail was sent, it will calculate the number of days until the allocation's expiration, and notify the PI of that specific number.

We persist this information in the allocations_allocation table with the new field 'expiration_warning_issued', which is a timestamp (default NULL). PIs will only be sent an email if this value is NULL. This means that the only anticipated circumstance under which someone will receive two warning emails is if there is an error writing the timestamp of their previous email to the database.


allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+from django.core.mail import send_mail
 from django.db import transaction
+from django.utils.html import strip_tags
 from allocations.models import Allocation
 from allocations.allocations_api import BalanceServiceClient
 from util.keycloak_client import KeycloakClient
 
 LOG = logging.getLogger(__name__)
 
+def warn_user_for_expiring_allocation():
```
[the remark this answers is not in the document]
super-cooper: Not consistent with the style of the rest of the repo

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+    now = datetime.now()
+    today = now.date()
+    days_this_month = calendar.monthrange(today.year, today.month)[1]
+    target_expiration_date = today + timedelta(days=days_this_month)
+
+    # Find allocations that are expiring between today and 1 month, and have not been warned of their impending doom
+    expiring_allocations = Allocation.objects.filter(status='active',
+                                                     expiration_date__lte=target_expiration_date,
+                                                     expiration_date__gte=today,
+                                                     expiration_warning_issued__isnull=True)
```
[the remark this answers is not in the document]
super-cooper: Done

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+
+        if expiration_date == today:
+            time_description = "today"
+        else:
+            # "in 1 day" | "in <n> days"
+            time_description = f"in {time_until_expiration.days} day{'s' if time_until_expiration.days != 1 else ''}"
+
+        email_body = f"""
+        <p>
+            The allocation for project <b>{charge_code}</b> will expire <b>{time_description}.</b>
```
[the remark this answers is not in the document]
super-cooper: Done

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+        if expiration_date == today:
+            time_description = "today"
+        else:
+            # "in 1 day" | "in <n> days"
+            time_description = f"in {time_until_expiration.days} day{'s' if time_until_expiration.days != 1 else ''}"
+
+        email_body = f"""
+        <p>
+            The allocation for project <b>{charge_code}</b> will expire <b>{time_description}.</b> 
+            See our
```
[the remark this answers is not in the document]
super-cooper: Done

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+            time_description = "today"
+        else:
+            # "in 1 day" | "in <n> days"
+            time_description = f"in {time_until_expiration.days} day{'s' if time_until_expiration.days != 1 else ''}"
+
+        email_body = f"""
+        <p>
+            The allocation for project <b>{charge_code}</b> will expire <b>{time_description}.</b> 
+            See our 
+            <a href="https://chameleoncloud.readthedocs.io/en/latest/user/project.html#recharge-or-extend-your-allocation">Documentation</a>
```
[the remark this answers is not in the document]
super-cooper: Done

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+            )
+        except Exception:
+            mail_sent = False
+            LOG.exception(mail_error)
+
+        # The database transaction is in a separate block of code from sending the mail itself
+        # so that we may differentiate between exceptions thrown via sending mail and via updating the database
+        if mail_sent:
+            emails_sent += 1
+            LOG.info(f"Warned PI about allocation {alloc.id} expiring on {alloc.expiration_date}")
```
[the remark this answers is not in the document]
super-cooper: ?

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+            if alloc.expiration_warning_issued:
+                LOG.warning(f"Issued a duplicate expiration warning for project {charge_code}.")
```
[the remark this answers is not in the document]
super-cooper: ?

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+            try:
+                with transaction.atomic():
+                    alloc.expiration_warning_issued = now
+                    alloc.save()
+            except Exception:
+                LOG.error(f"Failed to update ORM with expiration warning timestamp for project {charge_code}.")
```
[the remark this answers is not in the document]
super-cooper: ?

allocations/tasks.py
```diff
@@ -1,18 +1,91 @@
...
+        else:
+            LOG.error(mail_error)
+
+    LOG.debug(f"Needed to send mail for {len(expiring_allocations)}, and {emails_sent} emails were actually sent.")
```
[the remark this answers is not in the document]
super-cooper: ?

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
+    Sends an email to users when their allocation is within one month of expiring
+    """
+    now = datetime.now()
+    today = now.date()
+    days_this_month = calendar.monthrange(today.year, today.month)[1]
+    target_expiration_date = today + timedelta(days=days_this_month)
+
+    # Find allocations that are expiring between today and 1 month, and have not been warned of their impending doom
+    expiring_allocations = Allocation.objects.filter(
+        status='active',
```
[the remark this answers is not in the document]
diurnalist: Is there a constant defined for this?

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
+        expiration_date__lte=target_expiration_date,
+        expiration_date__gte=today,
+        expiration_warning_issued__isnull=True
+    )
+    emails_sent = 0
+    for alloc in expiring_allocations:
+        charge_code = alloc.project.charge_code
+        expiration_date = alloc.expiration_date.date()
```
diurnalist, line 37: .date() just grabs the date component (and drops the time component), yes?

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
 from allocations.allocations_api import BalanceServiceClient
 from util.keycloak_client import KeycloakClient
 
 LOG = logging.getLogger(__name__)
 
+def warn_user_for_expiring_allocation():
+    """
+    Sends an email to users when their allocation is within one month of expiring
+    """
+    now = datetime.now()
```
diurnalist, line 22: Watch out for timezone. I always forget exactly how to do it, but it's good to explicitly set the timezone to UTC before taking the date part out, because it can affect the date.

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
+        expiration_date__gte=today,
+        expiration_warning_issued__isnull=True
+    )
+    emails_sent = 0
+    for alloc in expiring_allocations:
+        charge_code = alloc.project.charge_code
+        expiration_date = alloc.expiration_date.date()
+
+        if not expiration_date:
+            LOG.debug(f"Allocation {alloc.id} has no expiration date.")
```
diurnalist, line 40: This might be worth a warning, as it's really unexpected behavior.

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
+            on how to recharge or extend your allocation.
+        </p>
+        """
+
+        mail_error = f"Failed to send expiration warning email for project {charge_code} to {alloc.project.pi.email}"
+        try:
+            mail_sent = send_mail(
+                subject=f"NOTICE: Your allocation for project {charge_code} expires {time_description}!",
+                from_email=settings.DEFAULT_FROM_EMAIL,
+                recipient_list=[alloc.project.pi.email],
```
diurnalist, line 65: IIRC there are possibly a few users who don't have the email attribute set--I think we should check for that here and issue a warning if so.

allocations/tasks.py
```diff
@@ -1,18 +1,93 @@
...
+            LOG.info(f"Warned PI about allocation {alloc.id} expiring on {alloc.expiration_date}")
+            if alloc.expiration_warning_issued:
+                LOG.warning(f"Issued a duplicate expiration warning for project {charge_code}.")
+            try:
+                with transaction.atomic():
+                    alloc.expiration_warning_issued = now
+                    alloc.save()
+            except Exception:
+                LOG.error(f"Failed to update ORM with expiration warning timestamp for project {charge_code}.")
+        else:
```
diurnalist, line 86: Rather than having this big nested if, I think an early return can be better:

```python
if not mail_sent:
  LOG.error(mail_error)
  continue
```

diurnalist: And, given how many branches there are now, it probably makes sense to pull the logic that process a single allocation into its own function, so the "early returns" can actually be expressed with `return` as opposed to `continue`.

diurnalist (commented):
I think this is basically good to go! Wanted to ask a question about the timezone thing, what do you think about setting it to UTC? There is a general assumption that we use UTC everywhere.

Great job, there's some fancy Django in here.

fancy

allocations/tasks.py
diurnalist: Yes, I think you can set the configuration in settings.py to set the interval to something like 24 hours, or maybe on a calendar cron-like syntax?

https://stackoverflow.com/questions/32449845/how-to-run-a-django-celery-task-every-6am-and-6pm-daily

allocations/tasks.py
```diff
@@ -1,18 +1,137 @@
...
+        charge_code = alloc.project.charge_code
+
+        # If we successfully sent mail, log it in the database
+        if mail_sent:
+            emails_sent += 1
+            LOG.info(
+                f"Warned PI about allocation {alloc.id} "
+                f"expiring on {alloc.expiration_date}"
+            )
+            if alloc.expiration_warning_issued:
```
diurnalist, line 110: Instead of doing this, I think an `assert alloc.expiration_warning_issued is None` is warranted in the _send_expiration_warning_mail. Asserts are good for things that the code should already be ensuring, but you're just making sure. Asserts fire only if a bad code change was introduced; they're a good circuit breaker before something critical gets called w/ bad input.

diurnalist (approved):
See note about the assert 👇🏻   -- otherwise looks great!

Re: the black autoformatting, we started rolling this out gradually, so if you're in a file that hasn't been touched in a bit, black will suggest things. We do want the entire repo to conform to the black styleguide in the future, so just because other parts of the repo don't match, doesn't mean we don't want to apply the suggested formattings.

allocations/tasks.py
super-cooper: I changed the interval. I think if it's only run once per day, it's ok if we have repeat warnings in our logs about projects with no PI email. How do you feel about that?

allocations/tasks.py
super-cooper: Good call!

super-cooper (commented):
Got your changes in. Hopefully good to go now :D

allocations/tasks.py
diurnalist: Yep, sounds good to me :)

diurnalist (approved):
Great work!

PR #219, opened 2021-07-16
Remove legacy account warning banner from dashboard

super-cooper:
This banner seems to be confusing to users, and is perhaps misleading.


diurnalist (changes requested):
Can you also remove the code that sets `show_migration_info`? I think this is probably the only place it's used.

diurnalist (approved):
Thanks!

PR #228, opened 2021-08-26
Add check to ignore reminder email for outage starting within one day

Mark-Powers:
When sending outage reminders, `reminder_sent` is checked for null. When an outage is saved that is scheduled for
within 1 day, we set this to the current date to avoid the task sending an email in the future.

### DIFF
--- a/chameleon_mailman/tasks.py
+++ b/chameleon_mailman/tasks.py
@@ -35,15 +35,18 @@ def send_outage_reminders(crontab_frequency, send_outage_reminder_before):
 
             subject = "Outage Reminder: {}".format(outage.title)
             body = "<b>Outage Start:</b> {}<br /><br />".format(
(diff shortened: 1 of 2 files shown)


chameleon_mailman/tasks.py
```diff
@@ -25,19 +25,28 @@ def send_outage_reminders(crontab_frequency, send_outage_reminder_before):
         seconds=(crontab_frequency * 60))
 
     for outage in upcoming_outages.all():
+        # Check for an outage scheduled close to start date
+        if outage.start_date < outage.created + timedelta(days=1):
+            outage.reminder_sent = outage.created
+            outage.save()
+            continue
```
diurnalist, line 32: Super clean!

diurnalist (approved):

PR #233, opened 2021-09-02
Add aldryn-bootstrap3 fork for django 3.2

Mark-Powers:
This creates a temporary fix for pages using the old aldryn-bootstrap3
plugin. We should install an updated boostrap plugin, migrate these
pages to that, and then delete the aldryn-bootstrap3 plugin. 

We can see if there are pages using this plugin with
`./manage.py cms list plugins`
before removing the plugin eventually.


diurnalist:
Vendoring these dependencies blows up the repo massively. Is it possible instead to make a public GitHub fork of those libraries, make the changes we need, and then update the Dockerfile to have like

```dockerfile
pip install git+https://github.com/chameleoncloud/aldryn-bootstrap3.git
```
or similar?

Mark-Powers:
Moved to #234

PR #234, opened 2021-09-03
Add legacy plugins

Mark-Powers:
This adds support for the legacy plugin forks, which can be used to remove the plugins from the few pages they exist on. Bootstrap4 can replace the Bootstrap3 plugins, and we can delete the single teaser instance.


chameleon/settings.py
```diff
@@ -150,13 +152,28 @@
     # djangocms-blog
     "filer",
     "easy_thumbnails",
+    "mptt",
```
diurnalist, line 155: what is this for?

diurnalist (approved):
Much easier to review this one :)

chameleon/settings.py
Mark-Powers: It's a dependency of Filer, which is a dependency of bootstrap 4: https://django-filer.readthedocs.io/en/latest/installation.html

PR #235, opened 2021-09-03
Remove legacy plugins

Mark-Powers:
I've updated the cms pages that were using these, and now they can be removed.


diurnalist (approved):
What about that mptt thing, was that a dependency on bootstrap4?

PR #239, opened 2021-09-07
Fix sharing portal project sharing and filter deleted projects

Mark-Powers:
This fixes `'ShareTarget' object has no attribute 'charge_code'` error when an artifact was shared with a project.

It also filters deleted artifacts. There is no UI to do this, but an admin can set this via the admin panel.

### DIFF
--- a/sharing_portal/views.py
+++ b/sharing_portal/views.py
@@ -85,7 +85,7 @@ def check_view_permission(func):
                     for p in mapper.get_user_projects(
                         request.user.username, fetch_balance=False)
                 ]
-                if any(p.charge_code in user_projects for p in project_shares):
+                if any(p.project.charge_code in user_projects for p in project_shares):
                     return all_versions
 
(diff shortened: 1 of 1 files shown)


sharing_portal/views.py
```diff
@@ -60,6 +60,9 @@ def wrapper(request, *args, **kwargs):
 
 def check_view_permission(func):
     def can_view(request, artifact):
+        if artifact.deleted:
```
diurnalist, line 63: If deleted artifacts are filtered out at query time (which I think is the right way to do it), is this necessary?

diurnalist (approved):
Thanks! Feel free to merge after seeing my comments.

sharing_portal/views.py
Mark-Powers: @diurnalist Yes, as without this check, you could still view the page/launch an artifact that has been deleted.

PR #240, opened 2021-09-08
allow pi to manage roles and allow managers to manage membership

zhenz:


projects/pub_views.py
```diff
@@ -13,6 +13,14 @@
 logger = logging.getLogger("projects")
 
 
+def can_add_publications(user, project):
+    if is_admin_or_superuser(user):
+        return True
+    if user.username == project.pi.username:
```
diurnalist, line 19: Would "user == project.pi" also work?

projects/views.py
```diff
@@ -68,6 +68,34 @@ def project_member_or_admin_or_superuser(user, project, project_user):
     return False
 
 
+def can_manage_project_membership(user, project):
```
diurnalist, line 71: These three functions mean 3 calls to Keycloak will happen whenever this page loads. I think we should instead have a general function to get the scopes/roles for the user/group combination, and then set the flags in the context like this:

```
"can_manage_project": "manager" in scopes,
"can_manage_project_membership": "manager" in scopes,
```

projects/views.py
```diff
@@ -122,7 +152,11 @@ def view_project(request, project_id):
     type_form = EditTypeForm(**type_form_args)
     pubs_form = AddBibtexPublicationForm()
 
-    if request.POST and project_pi_or_admin_or_superuser(request.user, project):
+    if (
+        request.POST
+        and can_manage_project_membership(request.user, project)
```
diurnalist, line 157: Nice job checking this on the POST as well :)

diurnalist (commented):
From our discussion I think you have more UI changes coming in here, but I left some initial comments, it's looking great!

projects/templates/projects/view_project.html
```diff
@@ -133,43 +133,92 @@ <h3>Project Users</h3>
...
+  <tr>
+    <td>{{ u.username }}</td>
+    <td>{{ u.first_name }}</td>
+    <td>{{ u.last_name }}</td>
+    <td>{{ u.email }}</td>
+    <td>
+       <form role="form" method="post" style="display:inline">
+         {% csrf_token %}
+         <input type="hidden" name="user_ref" value="{{ u.username }}">
+         <select id="role_{{ u.email }}" name="user_role" class="form-select role_selector" disabled>
```
diurnalist, line 157: I think the user ID is probably safer to put here than the email, from an encoding (and XSS injection) standpoint.

projects/templates/projects/view_project.html
```diff
@@ -133,43 +133,92 @@ <h3>Project Users</h3>
...
+           {% endfor %}
+         </select>
+         <button type="submit" id="submit_role_{{ u.email }}" class="btn btn-xs btn-primary" name="change_role" style="display:none;">Submit</button>
+         <button type="button" onclick="resetRole('role_{{ u.email }}')" id="cancel_role_{{ u.email }}" class="btn btn-xs btn-danger" style="display:none;">Cancel</button>
+       </form>
+    </td>
+    <td>
+      {% if u.username != request.user.username %}
+      <div class="dropdown">
+        <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" {% if not can_manage_project and not can_manage_project_membership %}disabled{%endif%}>
```
diurnalist, line 169: What is the difference b/w `can_manage_project` and `can_manage_project_membership`? Given that this button is for managing membership, is it enough to check for the presence of that permission?

projects/templates/projects/view_project.html
```diff
@@ -133,43 +133,92 @@ <h3>Project Users</h3>
...
+          <i class="fa fa-gear text-info dropbtn"></i><span class="sr-only">Manager Member</span>
+        </button>
+        <div class="dropdown-menu dropdown-content" aria-labelledby="dropdownMenuButton">
+          {% if can_manage_project %}
+          <button onclick="enableChangeRole('{{ u.email }}')" class="btn btn-xs btn-block">Change Role</button>
```
diurnalist, line 174: Related to the above, how can somebody manage the project but not the membership? It would seem that either they can manage membership, in which case we should the button and form, or they can't, in which case we should show neither.

projects/templates/projects/view_project.html
```diff
@@ -133,43 +133,92 @@ <h3>Project Users</h3>
...
+  <tr>
+    <td>--</td>
+    <td>--</td>
+    <td>--</td>
+    <td>{{ i.email_address }}</td>
+    <td>Pending Invitation - {{ i.status }}</td>
+    <td>
+    <div class="dropdown">
+      <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" {% if not can_manage_project and not can_manage_project_membership %}disabled{%endif%}>
+        <i class="fa fa-gear text-info dropbtn"></i><span class="sr-only">Manager Member</span>
```
diurnalist, line 199: What does "Manager Member" mean in this span?

projects/views.py
```diff
@@ -122,7 +133,15 @@ def view_project(request, project_id):
     type_form = EditTypeForm(**type_form_args)
     pubs_form = AddBibtexPublicationForm()
 
-    if request.POST and project_pi_or_admin_or_superuser(request.user, project):
+    role, scopes = get_user_project_role_scopes(keycloak_client, request.user, project)
+    can_manage_project_membership = "manage-membership" in scopes
```
diurnalist, line 137: 👍🏻

util/keycloak_client.py
```diff
@@ -2,6 +2,7 @@
 from keycloak.admin.users import User, Users
 from keycloak.admin.groups import Groups
 from keycloak.admin.user.usergroup import UserGroups
+from keycloak.admin.user.usergrouproles import UserGroupRoles
```
diurnalist, line 5: Do we need to bump a dependency somewhere in the setup.py or poetry.lock to pull in this new version of the keycloak client?

projects/static/projects/js/members.js
```diff
@@ -0,0 +1,28 @@
+function resetRole(selector_id) {
```
diurnalist, line 1: nit: in Javascript, snake_case is almost never used, rather, it would be camelCase.

projects/static/projects/js/members.js
```diff
@@ -0,0 +1,28 @@
...
+  var selector = document.getElementById(selector_id);
+  for (var j = 0; j < selector.options.length; j++) {
+    var opt = selector.options[j];
+    if (opt.disabled) {
+          selector.value = opt.value;
+          break;
+    }
+  }
+  selector.disabled = true;
+  document.getElementById("submit_" + selector.id).style.display = "none";
```
diurnalist, line 11:
```suggestion
  document.getElementById("submit_" + selector_id).style.display = "none";
```
selector_id is already there, it doesn't need to get it from the selector itself, and this could prevent a bug if the id attribute goes away, or selector becomes a different type at somepoint, etc. Best to not add more constraints than necessary on the "selector" interface.

projects/static/projects/js/members.js
```diff
@@ -0,0 +1,28 @@
+function resetRole(selector_id) {
+  var selector = document.getElementById(selector_id);
+  for (var j = 0; j < selector.options.length; j++) {
+    var opt = selector.options[j];
+    if (opt.disabled) {
```
diurnalist, line 5: In front-end land, it is considered a bit of a no-no to do this, relying on presentation of the UI as a holder of state. JS has data attributes now, which you could use to store the initial/current role state in order to reset it.

```html
<select data-initial="member">
  <option value="member">Member</option>
</select>
```

```js
document.querySelect('select')[0].dataset.initial
```

The reason for watching out for storing some "state" in the disabled attribute is, what if some other UI element decides to disable this later for whatever reason? Or what if there is some other weird browser thing that temporarily disables the input? There are tons of ways an element could get disabled, not all of which (indeed, not most of which) are under this code's control.

projects/static/projects/js/members.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function enableChangeRole(email) {
+  var elem_id = "role_" + email;
+  var selectors = document.getElementsByClassName("role_selector");
+  for (var i = 0; i < selectors.length; i++) {
+    var selector = selectors[i];
+    if (selector.id == elem_id) {
+      selector.disabled = false;
+      document.getElementById("submit_" + selector.id).style.display = "inline";
```
diurnalist, line 22: I think if you set it to "", it effectively resets it to its old value, which is a bit more friendly to the CSS on the page.

diurnalist (approved):
Looks good! I left some comments from my days of frontend JS pain :)

projects/templates/projects/view_project.html
zhenz: admins "can manage project", including membership. managers "can only manage membership". (i.e. they can only add/remove users, but can't change roles of the members.) that is to say, only PI can promote/demote members.

Mark-Powers (approved):
Looks good to me! I only have one question:
I didn't check the keycloak repo carefully, but if `can_manage_project = True`, does that mean `can_manage_project_membership = True`? I see in the code for updating membership, only only `can_manage_project_membership` is checked, and I imagine that PIs would want managers who can edit project settings, and membership at the same time.

super-cooper (approved):
lgtm

zhenz:
> Looks good to me! I only have one question:
> I didn't check the keycloak repo carefully, but if `can_manage_project = True`, does that mean `can_manage_project_membership = True`? I see in the code for updating membership, only only `can_manage_project_membership` is checked, and I imagine that PIs would want managers who can edit project settings, and membership at the same time.

Yes, if someone can manage project meaning they can manage membership. For project settings, do you mean recharging/renewing allocations? I think that's a good point! in this commit, managers can only add/remove users/invitations. I think I'll leave it as it is for now. And we can enable more responsibilities for managers later.

PR #244, opened 2021-09-23
Allow deleting entire Trovi artifact

super-cooper:
Adds a button to delete all artifact versions, and then the artifact itself. The button is accessible via the edit page for a given artifact. Currently, it will launch a JavaScript prompt that will ask for the name of the artifact to confirm its deletion. It will wait until the correct name is input, or the prompt is cancelled.

This achieves the desired function of the confirmation prompt, but I think it would be more ideal to mimic the behavior of other websites (such as deleting a repo on GitHub) where a prompt pops up, and the confirm button is greyed out until the text in the box matches the desired input. I think this is more familiar to users, but unfortunately seems like a challenging front end task, which I am not very familiar with.

Some additional refactoring to the version deletion code has been done on the backend as well, which should make error reporting for delete_version and delete_all more user-friendly.

image

image

image

Button is disabled if artifact has DOI:
image

### DIFF
--- a/sharing_portal/templates/sharing_portal/edit.html
+++ b/sharing_portal/templates/sharing_portal/edit.html
@@ -37,11 +37,16 @@
 
   <div class="layoutGrid__side">
     <div class="artifactVersions">
-      <div style="display: flex">
-        <h4 style="flex: 1"><i class="fa fa-files-o"></i> Versions</h4>
-        <form role="form" method="post" style="display:inline; margin-right: 6%">
+      <div class="artifactVersions__Header">
+        <h4><i class="fa fa-files-o"></i> Versions</h4>
+        <form role="form" method="post">
           {% csrf_token %}
-          <button class="btn btn-danger" type="submit" name="delete_all" >Delete All</button>
+          <button class="btn btn-xs btn-danger" type="submit" name="delete_all"
+            {% if artifact.doi %}disabled title="Cannot delete artifact published with DOI"
+            {% else %}title="Delete the entire Trovi artifact"
+            {% endif %}>
+            Delete All
+          </button>
         </form>
       </div>
       <form method="post">
--- a/sharing_portal/templates/sharing_portal/edit.html
+++ b/sharing_portal/templates/sharing_portal/edit.html
@@ -37,7 +37,13 @@
 
   <div class="layoutGrid__side">
     <div class="artifactVersions">
-      <h4><i class="fa fa-files-o"></i> Versions</h4>
+      <div style="display: flex">
+        <h4 style="flex: 1"><i class="fa fa-files-o"></i> Versions</h4>
+        <form role="form" method="post" style="display:inline; margin-right: 6%">
+          {% csrf_token %}
+          <button class="btn btn-danger" type="submit" name="delete_all" >Delete All</button>
+        </form>
+      </div>
       <form method="post">
         {% csrf_token %}
         <ol>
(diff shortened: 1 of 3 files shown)


sharing_portal/views.py
```diff
@@ -215,19 +241,49 @@ def edit_artifact(request, artifact):
...
+
+        elif "delete_all" in request.POST:
+            for version in artifact.versions:
+                if not _delete_artifact_version(request, version):
+                    messages.add_message(
+                        request,
+                        messages.ERROR,
+                        "Could not delete all artifact versions.",
+                    )
+                    return HttpResponseRedirect(
```
Mark-Powers, line 264: Should it attempt to delete all versions before erroring? If one version has a doi, but the others don't, this will stop at the first version, but it could continue.

Mark-Powers (approved):
Looks great! Another thing that is open for discussion is whether to use inline css styles vs. putting them in a static file. Most of the time, I prefer to avoid inline css styles.

sharing_portal/views.py
super-cooper: I think regardless, we don't want to delete artifacts with undeleted versions, so it shouldn't matter if it goes all the way or not, right? Or maybe, we should allow deleting versions with DOI if we're wanting to delete the entire artifact anyway? I'm not too familiar with DOI, or the implications of deleting an artifact which is linked to one.

diurnalist: I think the "delete all" button probably shouldn't even be there if any version has a DOI; deleting will be impossible necessarily.

sharing_portal/templates/sharing_portal/edit.html
```diff
@@ -70,6 +76,20 @@ <h4><i class="fa fa-files-o"></i> Versions</h4>
                 event.preventDefault();
               }
             });
+          })
+          const deleteAllButton = document.querySelector('[name=delete_all]');
+          deleteAllButton.addEventListener('click', (event) => {
+            let warning = "WARNING! You are about to delete your entire Trovi artifact! " +
+              "Are you sure you want to do this? Type the name of the artifact to confirm:";
+            while (true) {
```
diurnalist, line 84: I don't think you need a while: true here; the prompt should be blocking. In fact, won't this mean the thing keeps coming up if the user tries to cancel via another mechanism?

sharing_portal/views.py
```diff
@@ -205,6 +205,32 @@ def _fetch_artifacts(filters):
...
 
+def _delete_artifact_version(request, version):
+    if not version.doi:
+        try:
+            version.delete()
+        except Exception as e:
+            messages.add_message(
+                request,
+                messages.ERROR,
+                f"Error deleting artifact version {version}: {str(e)}",
```
diurnalist, line 216: In general we should not be showing exception text to end-users. I think it is best to log the exception internally and show them a generic error.

sharing_portal/templates/sharing_portal/edit.html
```diff
@@ -37,7 +37,13 @@ <h2 class="artifactTitle">Edit artifact</h2>
 
   <div class="layoutGrid__side">
     <div class="artifactVersions">
-      <h4><i class="fa fa-files-o"></i> Versions</h4>
+      <div style="display: flex">
```
diurnalist, line 40: Please give this a class and do this in CSS as opposed to inline styles. Inline styles can get out of hand and have a habit of building on themselves over time.

sharing_portal/views.py
diurnalist: related to @Mark-Powers' comment, another arguably cleaner way to do this, since you're returning a bool from that function, is this:

```python
if not all(_delete_artifact_version(request, v) for v in artifact.versions):
    # handle not all of them deleting properly
```

That will attempt to delete any that it can.

sharing_portal/templates/sharing_portal/edit.html
```diff
@@ -37,7 +37,13 @@ <h2 class="artifactTitle">Edit artifact</h2>
...
+        <h4 style="flex: 1"><i class="fa fa-files-o"></i> Versions</h4>
+        <form role="form" method="post" style="display:inline; margin-right: 6%">
+          {% csrf_token %}
+          <button class="btn btn-danger" type="submit" name="delete_all" >Delete All</button>
```
diurnalist, line 44: I think adding `btn-xs` might make the button compete less visually.

diurnalist (changes requested):
Looking very good! I think we should either hide or visually disable the button if any of the versions have a DOI; in that event, it doesn't really make sense to "delete all", because we don't allow it.

zhenz (approved):
LGTM!

diurnalist (approved):
Great! Thanks for integrating all the changes. ⭐

PR #246, opened 2021-09-27
manage project fundings in portal

zhenz:


zhenz:
@diurnalist For importing existing fundings, I'd like to do it with migration by getting fundings from the allocation justification field and parse and save to the new Funding table. However, it's a lot trickier than importing from the google sheet, as the fundings were all free-form text in the justification field. For example, "This research is partially supported by NSF grant "<Title>" award # XXXXXX. It is also to be more completely funded by NSF Proposal "<Title>" proposal # XXXXXX." So I think I'll just create a one-time script and import things from the google sheet. which contains the parsed information.

@diurnalist @Mark-Powers and there will be conflicts with Mark's day pass changes. I'll address the conflicts once Mark's changes are ready to go.

projects/admin.py
```diff
@@ -1,6 +1,13 @@
 from django.contrib import admin
 
-from projects.models import Invitation, Publication
+from projects.models import Invitation, Publication, Funding
+
+
+def _get_charge_code(obj):
```
Mark-Powers, line 6: It would help to indicate what sort of object `obj` is, since this is in `projects/`, I at first assumed it is to get a project charge code.

Mark-Powers: Maybe just a comment would suffice

projects/static/projects/js/fundings.js
```diff
@@ -0,0 +1,50 @@
+function updateElementIndex(el, ndx) {
+    var id_regex = new RegExp('(form-\\d+)');
+    var replacement = 'form-' + ndx;
+    if ($(el).attr("for")) $(el).attr("for", $(el).attr("for").replace(id_regex, replacement));
+    if (el.id) el.id = el.id.replace(id_regex, replacement);
+    if (el.name) el.name = el.name.replace(id_regex, replacement);
+}
+
+function cloneMore(selector) {
+    var newElement = $(selector).clone(true);
```
diurnalist, line 10: I was looking at how one is "supposed" to do this kind of thing in Django and [came across something familiar](https://stackoverflow.com/questions/501719/dynamically-adding-a-form-to-a-django-formset) -- I think it is good to include source URLs when code is used like this, just so it's clearer where it came from (and the source may have more details about the reason.)

I wish there was a better solution for this! But, I'm having trouble thinking of one. The fundamental thing this is doing seems correct, cloning the last and then updating all the form refs so that Django does the right thing on the POST.

projects/views.py
```diff
@@ -510,6 +516,29 @@ def set_ks_project_nickname(chargeCode, nickname):
...
+            funding = Funding(**funding_dict)
+            funding.project_id = project_id
+            funding.is_active = True
+            funding.save()
+            saved_fundings.append(funding_dict)
+    return saved_fundings
+
+
+def _remove_fundings(before_list, after_list):
+    after_id_list = [f["id"] for f in after_list if f.get("id")]
```
diurnalist, line 533: How does this actually remove fundings?

diurnalist (approved):
I think this generally looks good, but it would be nice to play with it on dev! Nice UI wrangling :)

PR #247, opened 2021-09-28
lazy assign admin role to the PI when loading project detail page

zhenz:


diurnalist (approved):
Not a great fix long term, why don't we simply lazy-create this group in Keycloak?

PR #253, opened 2021-10-11
Clean up docker-compose

super-cooper:
The docker-compose.yml file we use for development had a couple issues with it. The `mysql` service had some duplicated keys, and port mappings were not quote-wrapped, which might confuse some tools.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - ./media:/media
       - static:/static
     ports:
-      - 127.0.0.1:8890:${PORTAL_PORT}
+      - "127.0.0.1:8890:${PORTAL_PORT}"
     entrypoint: ["python3", "manage.py"]
     command: ["runserver", "0.0.0.0:${PORTAL_PORT}"]
     depends_on:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -15,21 +15,17 @@ services:
       - ./media:/media
       - static:/static
     ports:
-      - 127.0.0.1:8890:${PORTAL_PORT}
+      - "127.0.0.1:8890:${PORTAL_PORT}"
```
diurnalist, line 18: i always wondered, (why) are the quotes necessary here? i can't think of what aspect of this would break yaml parsing, but i often see this quoted.

super-cooper: > i always wondered, (why) are the quotes necessary here? i can't think of what aspect of this would break yaml parsing, but i often see this quoted.

I chalk it up to bad yaml parsing, but also yaml does apparently support a lot of syntax I've never seen, so maybe there is something there.

super-cooper:
> Any idea why the healthcheck doesn't work? It looks innocuous enough.

Oh, I'm sure it worked. The issue was that there were _two_ healthchecks :)

PR #261, opened 2021-11-19
manage allocation charge via portal

zhenz:
- create charge model
- redesign the allcoation section of the project dashboard
- allow checking charge details via a popup modal
- add balance service as Django app
- update the portal periodic task for activating and expiring allocations
- remove balance service client
- create period task for checking and alerting


zhenz:
Screen Shot 2021-11-18 at 7 59 03 PM
Screen Shot 2021-11-18 at 8 00 38 PM

super-cooper (changes requested):
Excellent change that is very welcome by the support team, and I'm sure users will love as well :)

The code in here is really great, but there are a few things that need to be addressed, such as some unsanitized SQL queries.

Mark-Powers (approved):
Looks good to me! I agree with Adam's comments on parameterizing the SQL queries.

balance_service/enforcement/usage_enforcement.py
```diff
@@ -0,0 +1,322 @@
...
+
+        for f in date_formats:
+            try:
+                date = datetime.datetime.strptime(date_string, f)
+                return date
+            except ValueError:
+                continue
+
+    def _convert_to_localtime(self, utctime):
+        utc = utctime.replace(tzinfo=pytz.UTC)
```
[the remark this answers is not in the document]
zhenz: using pytz to convert to aware datetime instead of naive. and then uses django's timezone library to convert to local time.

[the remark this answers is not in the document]
zhenz: pytz is used to convert from naive to aware datetime. I searched a bit, but didn't see django has any functions for that.

super-cooper (approved):
LGTM 🏃

balance_service/enforcement/usage_enforcement.py
```diff
@@ -0,0 +1,322 @@
...
+
+            # get the ongoing charnge
+            ongoing_charge = [c for c in reservation_charges if c.end_time > now]
+            # should have exactly one ongoing charge
+            if len(ongoing_charge) == 1:
+                ongoing_charge = ongoing_charge[0]
+                ongoing_charge.end_time = now
+                ongoing_charge.save()
+            else:
+                LOG.error(
```
diurnalist, line 244: Is it appropriate to raise an exception here, do you think? Should this just proceed w/o an error if there are multiple ongoing charges for the reservation somehow? Currently the function will proceed, which will just make the problem worse by continuing to add more ongoing charges.

balance_service/utils/openstack/keystone.py
```diff
@@ -0,0 +1,82 @@
...
+
+from balance_service import exceptions
+
+LOG = logging.getLogger(__name__)
+
+
+class KeystoneAPI:
+    def __init__(self, token, auth_url=None):
+        if not auth_url:
+            self.auth_url = settings.OPENSTACK_TACC_AUTH_URL
```
diurnalist, line 14: Err, I don't think that's a good idea. I think auth_url should be required and an error raised if it's None. I don't think there's a sane thing we can fall back to. There is no guarantee the user will exist at TACC anymore, if they never logged in there.

balance_service/utils/su_calculators.py
```diff
@@ -0,0 +1,54 @@
...
+            allocated_sus = active_allocation.su_allocated
+            used_sus = sum(
+                [get_used_sus(charge) for charge in active_allocation.charges.all()]
+            )
+        project_balances.append(
+            {
+                "id": project.id,
+                "charge_code": project.charge_code,
+                "used": used_sus,
+                "allocated": allocated_sus,
```
diurnalist, line 51: As mentioned earlier, I think this needs `encumbered` to be returned too. You should be able to adjust the get_used_sus to maybe return a tuple of (used, encumbered), where encumbered is the total charge for `(actual_end_time - now) * cost`

balance_service/views.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@require_http_methods(["GET"])
+@authenticate
+def batch_get_project_allocations(keystone_api, request):
+    projects = request.GET["projects"]
+    if not projects:
+        return HttpResponseBadRequest("No projects specified")
+    # Assume comma-separated
+    projects = [float(pid) for pid in projects.split(",")]
```
diurnalist, line 53: Why is this a float?

projects/forms.py
```diff
@@ -188,11 +188,16 @@ class AllocationCreateForm(forms.Form):
     justification = forms.CharField(
         label="Resource Justification",
         help_text=(
-            "Please provide an update on the use of your current allocation - any "
-            "success stories, or just a general update on the "
-            "progress of your research on Chameleon. This is helpful "
-            "for us as we communicate with NSF regarding the value Chameleon is "
-            "bringing to the research community."
+            "If this is the first allocation request for the project, "
```
diurnalist, line 191: Was this part of the PR originally? Or a separate change maybe

projects/static/projects/css/projects.css
```diff
@@ -1,2 +1,4 @@
-.allocation{border-bottom:1px solid #eee;margin-top:22px;margin-bottom:22px}.allocation:last-child{border-bottom:none}.allocation-inactive{display:none}.label-pending{background-color:#F0AD4E}.label-approved,.label-active{background-color:#74A93C}
+/*.allocation{border-bottom:1px solid #eee;margin-top:22px;margin-bottom:22px}.allocation:last-child{border-bottom:none}*/
```
diurnalist, line 1: How is this generated?

projects/templates/projects/allocation.html
```diff
@@ -1,82 +1,62 @@
...
-        <a href="{% url 'projects:renew_allocation' project.id alloc.id %}"
-           title="">
-          <i class="fa fa-refresh"></i> Resubmit allocation
-        </a>
+      {% endif %}
+      <button class="btn btn-xs btn-block" {% if alloc.status != 'Approved' and alloc.status != 'Active' and alloc.status != 'Inactive' %}disabled{%endif%}>
+        {% if alloc.status != 'Approved' and alloc.status != 'Active' and alloc.status != 'Inactive' %}
+        View Charges
+        {% else %}
+        <a class="open-charge" href="{% url 'projects:view_charge' alloc.id %}" data-charge-popup-url="{% url 'projects:view_charge' alloc.id %}">View Charge</a>
```
diurnalist, line 56: How does the charge get rendered inline? Is this something handled automagically by the modal JS helper?

diurnalist (changes requested):
Only thing I think we should change is the fact that "encumberance" has been removed, don't we still need that to prevent too many leases created in the future?

Otherwise looks great :) -- I left some comments on style and suggestions but the encumberance thing is the only blocker.

projects/static/projects/css/projects.css
zhenz: manually :worried:

diurnalist: :) -- it may be a simple thing to install a formatter so it looks a bit nicer, I assumed it was output by a tool!

projects/templates/projects/allocation.html
zhenz: a function to open up the modal. https://github.com/ChameleonCloud/portal/blob/a3587decb2abd1c86f3b9e80948077b3e6827f6a/projects/static/projects/js/allocations.js#L12-L19. Don't know if answers your questions?

diurnalist (approved):
OK looks great! ⭐  - I left some comments we can address now or later, mostly suggestions

zhenz:
rollout plan:
- create a portal user for openstack db (or make sure there is one); grant read-only access to blazar db (both sites)
- check access from portal to openstack db servers (both sites)
- add openstack db params to env file (TACC_OPENSTACK_DB_HOST etc.)
- deploy portal (it should run the migration and seed the charge table)
- add portal balance service url to blazar.conf as another external_service_endpoint
- deploy blazar  updates https://github.com/ChameleonCloud/blazar/pull/55/files

diurnalist (approved):

diurnalist (changes requested):
See comments on the Blazar change--I agree this is a more resilient way to roll out the new endpoints and is superior to running the two filters in parallel but using 1 for enforcement and 1 for viewing charges.

balance_service/views.py
```diff
@@ -0,0 +1,169 @@
...
+    except ue_exceptions.BillingError as e:
+        logger.exception(e)
+        if balance_service_version == 2:
+            check = e
+    except Exception as e:
+        logger.exception(e)
+        if balance_service_version == 2:
+            check = _unexpected_error()
+
+    # balance service v1
```
diurnalist, line 114: I would think we only want to run this logic if `balance_service_version == 1`, yes?

balance_service/views.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@csrf_exempt
+@require_http_methods(["POST"])
+@authenticate
+@transaction.atomic
+def check_create(keystone_api, request):
+    enforcer = usage_enforcement.UsageEnforcer(keystone_api)
+    balance_service = BalanceServiceClient()
+    return _process(
```
diurnalist, line 134: Nice, very clean :)

diurnalist (approved):

balance_service/views.py
zhenz: @diurnalist logic here is we run both new and old balance services no matter what `balance_service_version` is, but ignore (only log) v1 errors if `balance_service_version == 2` and ignore (only log) v2 errors if `balance_service_version == 1`.

PR #263, opened 2021-11-23
Rename Day Pass to daypass

Mark-Powers:
I've tested that everything works locally. My only concern for deploying to prod is renaming the database model from `DayPassRequest` to `DaypassRequest`. Django lowercases the model name, and so locally this was fine. We should be careful about this, though I believe the only requests are from me, generated by my demos. Firstly, we could just check the if the table name is lowercase on prod, just to be safe. If it is lowercase, we can leave the patch as as. Otherwise, the PR can be changed to either:

- Leave the model as `DayPassRequest`, since this isn't exposed to users
- Set `db_table` on the model to what the table is currently in prod


zhenz (approved):
:dizzy_face: LGTM!

super-cooper (approved):
LGTM :paintbrush:

sharing_portal/migrations/0016_day_pass_fields.py
```diff
@@ -32,7 +32,7 @@ class Migration(migrations.Migration):
             field=models.ForeignKey(null=True, on_delete=models.deletion.CASCADE, related_name='reproducibility_project', to='projects.project'),
         ),
         migrations.CreateModel(
-            name='DayPassRequest',
+            name='DaypassRequest',
```
diurnalist, line 35: In general historic migrations should be considered frozen. Is this pure documentation? Or, does changing the class name of the model actually also change the table name?

Mark-Powers: @diurnalist Changing the classname in the model would change the table name, if they differed in more than just case, since Django just lower cases the model name to get the table name. So in this case, it functions as documentation, and isn't technically needed.

There is a migration operation `RenameModel`, which can be used to change table names when a model changes. However, it doesn't seem to work in this situation, since it seems the table name doesn't change.

diurnalist: OK, I wasn't sure if Django did conversion from camelCase to snake_case, thought maybe it would think it was daypass versus day_pass 👍🏻

diurnalist (approved):

PR #265, opened 2021-12-14
display institution and country of the pi

zhenz:
we'd like to grant pi eligibility to international researchers,
but have more restrict rules to approve their allocation requests.
in order to easily tell if the project belongs to an international
pi, the allocation admin page will display the institution and country
of the pi. The information is collected from keycloak.

### DIFF
--- a/allocations/views.py
+++ b/allocations/views.py
@@ -5,6 +5,7 @@ from django.core import validators
 from django.core.exceptions import ValidationError
 from django.http import HttpResponse
 from django.contrib.auth.decorators import login_required, user_passes_test
+from util.keycloak_client import KeycloakClient
 from util.project_allocation_mapper import ProjectAllocationMapper
 
 import logging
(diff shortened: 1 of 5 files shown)


super-cooper (approved):
LGTM 🌍

Mark-Powers (approved):
👍

allocations/views.py
```diff
@@ -55,9 +56,22 @@ def get_all_alloc(request):
...
         mapper = ProjectAllocationMapper(request)
         resp = mapper.get_all_projects()
         logger.debug("Total projects: %s", len(resp))
+        for r in resp:
+            pi_attributes = user_attributes.get(r["pi"]["username"])
+            if pi_attributes:
+                institution = pi_attributes.get("affiliationInstitution")
+                country = pi_attributes.get("country")
+                r["pi"]["institution"] = (
+                    institution[0] if institution and len(institution) > 0 else None
```
diurnalist, line 70: another way to do this is to change the list to an iterator and call `next`. It's a little bit clunky still, I wish python had a way to .peek() on a list and have it return None if there was no next item.

```python
r["pi"]["institution"] = next(iter(instititution), None)
```

allocations/views.py
```diff
@@ -55,9 +56,22 @@ def get_all_alloc(request):
         json: dumps all data as serialized json.
     """
     try:
+        keycloak_client = KeycloakClient()
+        user_attributes = keycloak_client.get_all_users_attributes()
```
diurnalist, line 60: We only need the PI's user attributes here, why does this fetch for all users? Seems like that could be expensive.

diurnalist (approved):
In general approved, but wondering if the Keycloak call can be more targeted.

Nice text and additional UI!

allocations/views.py
zhenz: this is for the allocation admin page, which will load all projects/allocations, and we'll need the attributes for all the PIs.

zhenz: Ha! I see your point - getting users with the pi role. will fix

PR #266, opened 2022-01-07
fix for getting resource id

zhenz:
when creating a new charge record, a temporary resource id is given.
a periodic task finds the actual reservation id and replace the
temporary one. However, getting the reservation id based on
project_id, user_id, lease_start_date is not enough, so lease name
condition is added.

### DIFF
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -287,13 +287,14 @@ def _fill_charge_tmp_resource_ids():
     for region in charge_by_region.keys():
         db = utils.connect_to_region_db(region)
         for charge in charge_by_region[region]:
-            items = charge.resource_id.split("/")
+            items = charge.resource_id.split("/", maxsplit=4)
             project_id = items[1]
             user_id = items[2]
             lease_start_date = items[3]
+            name = items[4]
             resource_type = charge.resource_type
             resource_id = utils.get_resource_id(
-                db, project_id, user_id, lease_start_date, resource_type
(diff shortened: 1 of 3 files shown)


allocations/tasks.py
```diff
@@ -287,13 +287,14 @@ def _fill_charge_tmp_resource_ids():
     for region in charge_by_region.keys():
         db = utils.connect_to_region_db(region)
         for charge in charge_by_region[region]:
-            items = charge.resource_id.split("/")
+            items = charge.resource_id.split("/", maxsplit=4)
```
diurnalist, line 290: ah, so leases can have slashes in the name 👍🏻

diurnalist (commented):
What if the user creates a lease, then deletes it, then creates another for the same start date, with the same name?

balance_service/enforcement/usage_enforcement.py
```diff
@@ -166,6 +166,7 @@ def check_usage_against_allocation(self, data):
                     project_id=lease["project_id"],
                     user_id=lease["user_id"],
                     start_date=lease["start_date"],
+                    name=lease["name"],
```
diurnalist, line 169: The root problem is that we don't have reservation.id at this point I take it?

zhenz: yes. reservation id is assigned after the enforcement checks.

super-cooper (approved):

Mark-Powers (approved):
👍

PR #269, opened 2022-01-19
Add field on user when joining allocation for the first time

diurnalist:
We'd like to start tracking when users join their first allocation for usage reporting purposes. It helps to know how many users are actually part of an allocation, and what is the delta b/w when they signed up, and when they finally were added to an allocation (if at all.)

### DIFF
--- a/projects/views.py
+++ b/projects/views.py
@@ -16,10 +16,7 @@ from django.utils import timezone
 from django.core.validators import validate_email
 from django.http import (
     Http404,
-    HttpResponse,
-    HttpResponseBadRequest,
     HttpResponseForbidden,
-    HttpResponseNotAllowed,
     HttpResponseRedirect,
     JsonResponse,
 )
@@ -33,6 +30,7 @@ from util.project_allocation_mapper import ProjectAllocationMapper
 from util.keycloak_client import KeycloakClient
 
 from projects.serializer import ProjectExtrasJSONSerializer
+from . import membership
 
 from .forms import (
     AddBibtexPublicationForm,
@@ -144,7 +142,7 @@ def accept_invite_for_user(user, invitation, mapper):
         invitation.accept(user)
         project = mapper.get_project(invitation.project.id)
         user_ref = invitation.user_accepted.username
-        mapper.add_user_to_project(project, user_ref)
+        membership.add_user_to_project(project, user_ref)
         return True
     return False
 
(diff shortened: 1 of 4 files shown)


util/project_allocation_mapper.py
```diff
@@ -402,26 +402,6 @@ def contact_pi_via_rt(self, data):
         setattr(alloc, "decision_summary", f"RT ticket created with id: {ticket_id}")
         alloc.save()
 
-    def _update_user_membership(self, tas_project, user_ref, action=None):
```
diurnalist, line 405: I'm starting to remove functions from this big mapper module, because the mapper should go away soon. It was only relevant when we were in this in-between step during user federation migration. We have some more tech debt to address here, notably the fact that we still use TAS project representations internally everywhere 😭

projects/views.py
```diff
@@ -774,7 +767,13 @@ def edit_project(request):
 def edit_nickname(request, project_id):
     mapper = ProjectAllocationMapper(request)
     project = mapper.get_project(project_id)
-    if not project_pi_or_admin_or_superuser(request.user, project):
```
diurnalist, line 777: This was a bug; this function no longer exists

projects/views.py
```diff
@@ -16,10 +16,7 @@
 from django.core.validators import validate_email
 from django.http import (
     Http404,
-    HttpResponse,
```
diurnalist, line 19: Imports not used (IDE plugins are helpful for catching this type of thing.)

diurnalist:
We will need to backfill this attribute for existing users. For all users who are already part of a project, the simplest solution would be to just set this field to their join date. A more complicated solution would be to set it to the create date of the Keycloak group.

projects/views.py
```diff
@@ -261,10 +257,7 @@ def view_project(request, project_id):
                         )
                 except Exception:
```
super-cooper, line 258: Not sure if it's in the scope of this patch, but might be nice to have an additional catch for `User.DoesNotExist`. I imagine this could get frustrating with an easy-to-miss typo.

super-cooper (approved):
Looks good! I only had one suggestion, but unsure if it belongs in this patch. Although the impact is minuscule, so maybe worth throwing in.

Mark-Powers (approved):

projects/views.py
diurnalist: We already catch this above: https://github.com/ChameleonCloud/portal/pull/269/files#diff-cec7e4e74af6d0ee95b13c06b15ad41ec82fac8041923a65a5a58fc270669f3bL227

PR #271, opened 2022-01-21
allow super users to update project pi

zhenz:


zhenz:
Screen Shot 2022-01-21 at 3 19 24 PM
After clicking the icon next to the PI
Screen Shot 2022-01-21 at 3 19 42 PM

projects/static/projects/js/allocations.js
```diff
@@ -29,6 +29,12 @@
         $('#icon_show_type_edit').toggleClass('hidden');
         $('#project_type_edit_form').toggleClass('hidden');
       });
+      
+    $('#icon_show_pi_edit,#btn_pi_cancel').on('click', function(e) {
```
diurnalist, line 33: needs an extra space I think 🔎

projects/views.py
```diff
@@ -828,6 +837,52 @@ def edit_type(request, project_id):
...
+        messages.error(request, "Only the admin users can update project PI.")
+        return EditPIForm(request.POST)
+
+    form = EditPIForm(request.POST)
+    if form.is_valid():
+        project_pi_username = form.cleaned_data["pi_username"]
+        # try to update PI
+        try:
+            new_pi = User.objects.get(username=project_pi_username)
+            if new_pi.pi_eligibility().lower() != "eligible":
```
diurnalist, line 852: Might be worth making this a boolean `is_pi_eligible` or something?

projects/views.py
```diff
@@ -828,6 +837,52 @@ def edit_type(request, project_id):
...
+                raise UserNotPIEligible()
+            ProjectAllocationMapper.update_project_pi(project_id, project_pi_username)
+            form = EditPIForm(request.POST)
+            messages.success(
+                request,
+                "Update successful! Click <a href='{}'>here</a> to reload".format(
```
diurnalist, line 858: I don't _think_ this is vulnerable to injection, but the user may have control over `request.path`. Probably not much control as it must have some set of values to even be running in this function. Do we need this "click here to reload" thing at all, do you think?

diurnalist: (given that this is only available to super users it's also less of an attack vector IMO)

diurnalist (approved):

projects/forms.py
```diff
@@ -118,6 +118,16 @@ def __init__(self, *args, **kwargs):
             logger.error("Couldn't get type list.")
 
 
+class EditPIForm(forms.Form):
+    pi_username = forms.CharField(
+        label="",
+        max_length="100",
```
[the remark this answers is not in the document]
zhenz: max_length is required for charfield. the default value is 50. I thought some email addresses may go over 50?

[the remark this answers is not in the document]
diurnalist: I don't know how well-utilized this RFC is, but according to it, the max length is 320 for email addresses: https://datatracker.ietf.org/doc/html/rfc3696#section-3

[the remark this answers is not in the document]
zhenz: :astonished: :thumbsup:

PR #286, opened 2022-03-16
Add trovi migration task

Mark-Powers:
This task can be run by an operator via the shell. Some manual
intervention is required, as the script needs assistance matching author
email in the case there is more than one account with the given full
name.

Note: Some of the funtions in `trovi.py` are unused by this task, but will be 
used by the portal GUI.


sharing_portal/tasks.py
```diff
@@ -67,10 +72,186 @@ def _temp_url(deposition_id):
...
+        "op": op,
+        "path": f"/authors/{index}",
+        "value": new_json_author,
+    }
+
+
+def _get_trovi_token(client_credentials=False):
+    keycloak_client = KeycloakClient()
+    realm = KeycloakRealm(server_url=keycloak_client.server_url, realm_name="chameleon")
+    if client_credentials:
```
diurnalist, line 237: nice

sharing_portal/trovi.py
```diff
@@ -0,0 +1,273 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class TroviException(Exception):
+    pass
+
+
+def url_with_token(path, token):
+    req = requests.PreparedRequest()
```
diurnalist, line 22: I have seen this a few times in PRs but I don't really get it, is it just to make it slightly easier to put a query string on a URL? Django in particular has this:

```python
from django.http import QueryDict
from urllib.parse import urljoin

q = QueryDict(mutable=True)
q["access_token"] = token
url = f"{urljoin(os.getenv('TROVI_API_BASE_URL'), path)}?{q.urlencode()}"
```

And here's a non-Django way, even simpler if you aren't trying to support multi-value keys:

```python
from urllib.parse import urljoin, urlencode

q = {"access_token": token}
url = f"{urljoin(os.getenv('TROVI_API_BASE_URL'), path)}?{urlencode(q)}"
```

a PreparedRequest is to my knowledge mostly used if you need to do some low-level adjustments to the request or are using request sessions.

sharing_portal/trovi.py
```diff
@@ -0,0 +1,273 @@
...
+    req.prepare_url(
+        urljoin(os.getenv("TROVI_API_BASE_URL"), path), dict(access_token=token)
+    )
+    return req.url
+
+
+def check_status(response, code):
+    if response.status_code != code:
+        print(response.text)
```
diurnalist, line 31: Maybe should remove this?

sharing_portal/trovi.py
```diff
@@ -0,0 +1,273 @@
...
+    realm = KeycloakRealm(server_url=keycloak_client.server_url, realm_name="master")
+    openid = realm.open_id_connect(
+        client_id=keycloak_client.client_id, client_secret=keycloak_client.client_secret
+    )
+    creds = openid.client_credentials()
+    return get_token(creds["access_token"], is_admin=True)["access_token"]
+
+
+def get_token(token, is_admin=False):
+    scope = "artifacts:read artifacts:write"
```
diurnalist, line 62: I think lists make it harder to add bugs later:

```python
scopes = ["artifacts:read", "artifacts:write"]
if is_admin:
    scopes.append("trovi:admin")

# later...
"scope": " ".join(scopes)
```

sharing_portal/trovi.py
```diff
@@ -0,0 +1,273 @@
...
+
+def get_trovi_uuid(artifact_id):
+    artifact = Artifact.objects.get(pk=artifact_id)
+    return artifact.trovi_uuid
+
+
+def get_author(author, prompt_input=True):
+    matching_users = list(
+        User.objects.annotate(full_name=Concat("first_name", "last_name"))
+        .filter(full_name=author.name.replace(" ", ""))
```
diurnalist, line 108: heh, clever. You could also do this to avoid the space replacement:

```python
User.objects.annotate(full_name=Concat("first_name", Value(" "), "last_name"))
```

sharing_portal/trovi.py
```diff
@@ -0,0 +1,273 @@
...
+        email = selected_user.email
+    return {
+        "full_name": author.name,
+        "affiliation": author.affiliation,
+        "email": email,
+    }
+
+
+def portal_artifact_to_trovi(portal_artifact, prompt_input=True):
+    print(portal_artifact.title)
```
diurnalist, line 132: Maybe remove

diurnalist (approved):
I think overall this looks fine, some suggestions included. Maybe the case of updating existing could be simpler w/ jsonpatch. The reason for designing updates as jsonpatch in the API was supposed to be that it allowed clients to make their lives easier, as it's a pretty well-defined spec. Maybe that's not panning out though.

sharing_portal/trovi.py
diurnalist: I noticed that as well in the data. I think in most cases it was trailing or leading whitespace. Maybe `author.name.trim()`?

sharing_portal/trovi.py
diurnalist: In the requests lib the more common thing is to not build it at all, just fire it immediately by passing w/e args in. I think PreparedRequest is more commonly used if you need more advanced processing: https://docs.python-requests.org/en/latest/user/advanced/#prepared-requests

`requests.get` for example will construct a PreparedRequest under the hood, and then send it. In general the higher-level abstractions are a bit easier b/c you can also mock them easier for testing.

sharing_portal/tasks.py
```diff
@@ -67,10 +72,186 @@ def _temp_url(deposition_id):
...
+    if existing_trovi_uuid:
+        print(
+            f"Artifact already created on trovi {existing_trovi_uuid}, checking for updates"
+        )
+        artifact_in_trovi = trovi.get_artifact(token, artifact_id)
+        artifact_in_portal = trovi.portal_artifact_to_trovi(
+            Artifact.objects.get(pk=artifact_id),
+            prompt_input=False,
+        )
+        patches = []
```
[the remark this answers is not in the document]
Mark-Powers: yeah this is much better than what I was doing

sharing_portal/trovi.py
Mark-Powers: Sure, I can replace this code.

sharing_portal/trovi.py
```diff
@@ -0,0 +1,264 @@
...
+    chameleon_projects = [
+        lp for lp in artifact["linked_projects"] if lp.split(":", 4)[3] == "chameleon"
+    ]
+    if not chameleon_projects:
+        return None
+    charge_code = chameleon_projects[0].split(":", 4)[4]
+    return Project.objects.get(charge_code=charge_code)
+
+
+def set_linked_project(artifact, charge_code, token=None):
```
super-cooper, line 241: You should be able to do this in the create request if it has an admin token. Was it failing for you?

super-cooper (commented):
Left some comments on trovi.py. I will review the rest of it +updates to trovi.py next week.

sharing_portal/trovi.py
```diff
@@ -0,0 +1,264 @@
...
+    json_data = {"patch": patches}
+    print(json_data)
+    res = requests.patch(
+        url_with_token(f"/artifacts/{artifact_uuid}/", token), json=json_data
+    )
+    check_status(res, requests.codes.ok)
+    return res.json()
+
+
+def create_version(token, trovi_artifact_uuid, contents_urn, links=[]):
```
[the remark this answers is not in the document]
diurnalist: Good callout, I often forget that Python evaluates default args once, at module eval time.

sharing_portal/trovi.py
```diff
@@ -0,0 +1,264 @@
...
+                "contents": {
+                    "urn": f"urn:trovi:contents:{version.deposition_repo}:{version.deposition_id}"
+                },
+                "metrics": {"access_count": version.launch_count},
+                "links": [],
+            }
+            for version in portal_artifact.versions.all()
+        ]
+    else:
+        version = list(portal_artifact.versions.all())[0]
```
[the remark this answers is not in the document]
super-cooper: Huh, you're right. Whoops.

zhenz (approved):
in general looks good to me. It's a bit difficult to review the syncing script without details of how the new mechanism works.

super-cooper (approved):
The rest of this LGTM. I still need to put in a patch to trovi to let admins force-update the readonly fields, after which this will need some small changes to update those fields.

PR #290, opened 2022-03-21
Update sharing portal to use trovi as backend

Mark-Powers:


sharing_portal/forms.py
```diff
@@ -1,89 +1,59 @@
...
-        # Ensure only staff members can save w/ the "chameleon" label.
-        if (
-            any(l.label == Label.CHAMELEON_SUPPORTED for l in labels)
-            and not self.request.user.is_staff
-        ):
-            raise ValidationError("Invalid label")
-        return labels
+        available_labels = [
+            (t["tag"], t["tag"])
+            for t in trovi.list_tags(request.session.get("trovi_token"))
```
diurnalist, line 29: 👍🏻

sharing_portal/forms.py
```diff
@@ -175,19 +145,25 @@ def clean(self):
         return super(ZenodoPublishForm, self).clean()
 
     def _has_doi(self):
-        return self.model.doi is not None
+        return _version_is_zenodo(self.model)
+
+
+def _version_is_zenodo(version):
+    return version["contents"]["urn"].split(":", 2)[1] == "zenodo"
```
diurnalist, line 152: 👍🏻

sharing_portal/migrations/0018_auto_20220317_1314.py
```diff
@@ -0,0 +1,30 @@
...
+        migrations.RemoveField(
+            model_name='daypassrequest',
+            name='artifact',
+        ),
+        migrations.AddField(
+            model_name='daypassrequest',
+            name='artifact_uuid',
+            field=models.CharField(max_length=36, null=True),
+        ),
+        migrations.CreateModel(
```
diurnalist, line 23: How will we handle existing requests (and does it matter?) Normally you would do a dance like, add the new column, then do a data migration, then remove the old column.

sharing_portal/templates/sharing_portal/detail.html
```diff
@@ -23,28 +23,31 @@
         <h2 class="artifactTitle">{{ artifact.title }}</h2>
         <div class="artifactActions">
           {% if editable %}
-          <a class="btn btn-default" href="{% url 'sharing_portal:share' artifact.pk %}">
+          <a class="btn btn-default" href="{% url 'sharing_portal:share' artifact.uuid %}">
```
diurnalist, line 26: Very nice how easy this template was to update

sharing_portal/templates/sharing_portal/embed.html
```diff
@@ -22,10 +22,6 @@ <h2 class="artifactTitle">{{ form_title }}</h2>
           </div>
         </header>
 
-        {% if version_form %}
```
diurnalist, line 25: This whole template can go away, we won't need an embedded edit/create form anymore.

sharing_portal/templates/sharing_portal/share.html
```diff
@@ -86,27 +66,12 @@ <h2 class="artifactTitle">Edit sharing settings</h2>
...
-          {% endfor %}
-        </div>
       </div>
     </div>
   </form>
 
 </div>
 
-{# TODO: replace with json_script filter when we're on Django 2.1 #}
-<script id="artifact-data" type="application/json">{{ artifact_json|safe }}</script>
```
diurnalist, line 109: I forget what this was even for. I think maybe at some point I had an idea that we would have a Vue form or something, but that didn't work out.

sharing_portal/views.py
```diff
@@ -1,226 +1,180 @@
...
 
 
-class ArtifactSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Artifact
-        fields = "__all__"
+def ensure_trovi_token(view_func):
+    def _wrapped_view(request, *args, **kwargs):
+        if (
+            request.session.get("trovi_token_expiration")
```
diurnalist, line 48: These expirations are in UTC right?

sharing_portal/views.py
```diff
@@ -1,226 +1,180 @@
...
+            and datetime.now().timestamp() > request.session["trovi_token_expiration"]
+        ):
+            del request.session["trovi_token_expiration"]
+            del request.session["trovi_token"]
```
diurnalist, line 52: `.pop(key, None)` is safer as you avoid a potential KeyError, especially given that we did not check for "trovi_token" being on the session yet.

sharing_portal/views.py
```diff
@@ -1,226 +1,180 @@
...
+                        (
+                            datetime.now() + timedelta(seconds=response["expires_in"])
+                        ).timestamp(),
+                        request.session["oidc_token_expiration"],
+                    )
+                except trovi.TroviException:
+                    LOG.error("Error getting trovi token")
+            else:
+                # Set an empty token
+                request.session["trovi_token"] = ""
```
diurnalist, line 72: What is the effect of an empty token? The function implies it "ensures" a token, but it looks like all that is ensured is that the key is set on the session. Does it make sense to `raise`, or does that introduce too many hard-to-manage exception cases?

diurnalist: I guess what I'm saying is, I thought from the name of the function that it would raise if it could not ensure the token was there. It's fine if it's cleaner to not do that, but maybe just change the name to `with_trovi_token` or something

sharing_portal/views.py
```diff
@@ -1,226 +1,180 @@
...
+    return _wrapped_view
 
 
 def can_edit(request, artifact):
-    if artifact.created_by == request.user:
-        return True
-    if request.user.is_staff:
-        return True
-    return False
+    return _owns_artifact(request.user, artifact) or request.user.is_staff
```
diurnalist, line 79: Does the staff flag mean anything anymore? Does having staff status translate to portal being allowed to generate an admin token for them?

sharing_portal/views.py
```diff
@@ -1,226 +1,180 @@
...
     def wrapper(request, *args, **kwargs):
-        pk = kwargs.pop('pk')
-        artifact = get_object_or_404(Artifact, pk=pk)
+        pk = kwargs.pop("pk")
+        try:
+            artifact = trovi.get_artifact_by_trovi_uuid(
+                request.session.get("trovi_token"), pk
+            )
+        except trovi.TroviException:
+            raise Http404("That artifact does not exist, or is private")
```
diurnalist, line 90: Might be good to check trovi's response code to make sure it's 404. We would want to log 500s so that we know when trovi is being bad.

sharing_portal/views.py
```diff
@@ -360,55 +328,83 @@ def share_artifact(request, artifact):
...
     else:
+        # Use the first linked chameleon project
+        chameleon_projects = [
+            lp
+            for lp in artifact["linked_projects"]
+            if lp.split(":", 4)[3] == "chameleon"
+        ]
+        project = None
+        if chameleon_projects:
+            project = chameleon_projects[0].split(":", 4)[4]
```
diurnalist, line 383: Probably would be good to make URN extraction a helper in some utils function or something. I noticed the zenodo extraction part assumed the urn would have only 3 parts

diurnalist (approved):
I think this looks pretty good! I did not have time to inspect the views.py changes with great detail. If there is a particular function or thing you'd like me to look at closer, I'm happy to do so. I did a prelim overview of all those functions and the diffs looked good, but some of the daypass changes I found a bit complicated, notably what the artifact map is doing.

Can you summarize for which operations portal will use a service token (i.e., a token that is not admin but has elevated privileges, like for writing metrics), for which it will use admin token, and for which it will use the user's own token? That was the main thing I was looking at. I think it looks good, but an enumeration would be good for double-checking this.

Thanks!

sharing_portal/views.py
Mark-Powers: No, I've updated to utcnow.

sharing_portal/views.py
Mark-Powers: Fixed.

sharing_portal/views.py
Mark-Powers: This assumes someone with `is_staff` is also a trovi admin. I can change `with_trovi_token` so to use the service admin token if user is staff.

sharing_portal/templates/sharing_portal/detail.html
Mark-Powers: Yes, the templates were fairly smooth

sharing_portal/views.py
Mark-Powers: trovi just returns "seconds remaining" and so is timezone agnostic, but I switched to UTC to avoid weird daylight savings issues or something.

sharing_portal/views.py
Mark-Powers: Setting the token empty is set so the unauthenticated user has no token when talking to trovi.

I've changed this to `with_trovi_token` to make this clearer.

Mark-Powers (commented):
@diurnalist My biggest concern in views.py is `with_trovi_token`. I'm unsure what the best way to keep the trovi token valid is. I've noticed the current function doesn't 100% work, and sometimes it doesn't refresh correctly. Adam might be able to speak on this as well. Am I making incorrect assumptions about state?

> Can you summarize for which operations portal will use a service token (i.e., a token that is not admin but has elevated privileges, like for writing metrics), for which it will use admin token, and for which it will use the user's own token? That was the main thing I was looking at. I think it looks good, but an enumeration would be good for double-checking this.

The admin service token is used in the migrate task in the previous PR, the `increment_metric_count` (which needs to be updated based on the latest trovi PR), and setting the linked project. The rest is using the user's token. However, as mentioned in the comment, I can change it so the service token is always used for a staff member, allowing them to edit/share/view all trovi artifacts as the sharing portal works now.

super-cooper:
> Adam might be able to speak on this as well. Am I making incorrect assumptions about state?

My full review will be tomorrow but I saw my name so weighing in. Everything with regards to token acquisition looks fine at a glance. How are the tokens failing to refresh? Is the request to Trovi/token/ failing? Or is it that the tokens are expired when it seems they shouldn't be? If that's the case, one thing I notice that could be going wrong is that there is no "buffer" time between when the token expires and when it's refreshed. The `expires_in` variable is exact, and has the unfortunate property of being inaccurate because it takes time for the token to be sent over the network. 99.99% of the time this doesn't matter, but transient network and/or latency can make it so. 

I see that the session's expiration time is set based on `utcnow`, but the truth would be closer to the token's `iat` claim. It would be annoying to have to decode the token, and also annoying to try to predict when this latency occurs, so honestly, it might be a good idea to just set the session's expiration for about 10 seconds before it actually expires :shrug: if that's actually the issue

diurnalist:
@Mark-Powers thanks for the detailed breakdown. I don't think we need to worry about enabling admins to edit any artifact. TBH that was a bit confusing. We have user impersonation if we really need to do this, or we can generate an admin token by hand, or we can build something into the Django admin GUI (or implement such a GUI in Trovi itself, perhaps even better!)

I did notice the admin token was used for when there are too many daypass requests. I'm guessing that's because that is triggered by an end-user who does not have permissions on the artifact? I think it's fine, just calling out the other case I saw.

I will have a closer look at with_trovi_token. In general, this is how I thought it would work:

* A user logs in to portal via OpenID Connect. We store their access_token and refresh_token. There is _supposed_ to be middleware that is refreshing that access_token when it comes close to or has expired. The refresh token has a longer validity than the access_token (in Keycloak it is configured as "SSO Session Idle" I believe.)
* The trovi token has its own lifecycle, including expiration. I forget what we configured but all tokens have the same lifetime and it's pretty small. The intent is that you generally would always request a new token before doing some operation or set of operations.
* As an optimization, one could store the trovi token and its expiration and avoid always requesting new tokens if we still have a valid one. How useful this is depends on how often the token would be used versus how long it is valid for.

I suspect the root cause here is that the access_token in portal's session is not getting refreshed. I know this is not working as required, b/c we have that weird error where if you have a session on portal for a while, sometimes the hardware catalogue stops working. That indicates that the user's backing token has expired and wasn't refreshed.

sharing_portal/views.py
```diff
@@ -1,226 +1,186 @@
...
 
-class ArtifactSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Artifact
-        fields = "__all__"
+def with_trovi_token(view_func):
+    def _wrapped_view(request, *args, **kwargs):
+        if (
+            request.session.get("trovi_token_expiration")
+            and datetime.utcnow().timestamp() > request.session["trovi_token_expiration"]
```
super-cooper, line 49: I mentioned in a comment yesterday, but it might be good to give yourself some buffer room in case of weird timing causing the token to expire before trovi validates it.

sharing_portal/views.py
```diff
@@ -1,226 +1,186 @@
...
+def handle_get_artifact(request, uuid, sharing_key=None):
+    try:
+        return trovi.get_artifact_by_trovi_uuid(
+            request.session.get("trovi_token"), uuid
+        )
+    except trovi.TroviException as e:
+        if e.code == 404:
+            raise Http404("That artifact does not exist, or is private")
+        if e.code == 403:
+            raise PermissionDenied("You do not have permission to view that page")
```
super-cooper, line 91: I think here we should check if this 403 came from a token expiration. If that's the case, we can fetch a new token and retry.

sharing_portal/views.py
super-cooper: This is a good point. If we're only trying to access the NSS, it might be good to index with -1 instead.

sharing_portal/views.py
```diff
@@ -656,10 +687,18 @@ def list_daypass_requests(request, **kwargs):
         for project in keycloak_client.get_user_roles(request.user.username)
         if manage_membership_in_scope(project["scopes"])
     ]
+    trovi_artifacts = trovi.list_artifacts(request.session.get("trovi_token"))
```
super-cooper, line 690: It would be nice to one day be able to pass more filters to the ListArtifacts endpoint 😔

super-cooper (approved):
LGTM! I have a few comments on little things here and there. I'm delighted to see how smoothly this went :)

sharing_portal/trovi.py
```diff
@@ -57,7 +58,7 @@ def get_client_admin_token():
 
 
 def get_token(token, is_admin=False):
-    scopes = ["artifacts:read", "artifacts:write"]
+    scopes = ["artifacts:read", "artifacts:write"]#, "artifacts:write_metric"]
```
super-cooper, line 61: Did you mean to uncomment this?

sharing_portal/migrations/0018_auto_20220317_1314.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 3.2 on 2022-03-17 18:14
```
zhenz, line 1: maybe change the auto-generated file name to something meaningful?

zhenz (approved):
👍

PR #291, opened 2022-03-23
Store canned approval responses in portal

zhenz:
Currently, all canned pi eligibility and allocation review
responses are stored in a google doc and the responses are
copied and pasted everytime. We would like to have a better
centralized location for the canned responses. Since the
responses are used exclusively with portal, we'll use the
portal to store them. Also, add some javascript functions
for auto-filling based on the selections.


super-cooper (approved):
Neat :)

PR #294, opened 2022-03-31
Sync version dates in artifact migration script

super-cooper:
Simple fix which syncs the dates for versions to Trovi. I tested this and it works, but there were a couple versions which caused errors on the second run because they tried to add a version which already existed.


super-cooper:
Correct! I purged both portal and Trovi each time I tested. The second run goes fine, but there are errors on two artifacts as I mentioned. If something doesn't sync properly, we could manually fix it.

Mark-Powers:
Ah, I think that's fine. If you are using an old DB dump that will happen. I deleted the duplicate in the production database, as it didn't have any launches.

PR #302, opened 2022-04-15
Trovi download archive

Mark-Powers:
This hooks up Trovi's contents endpoint to the side panel on the detail page for artifacts.

Here is an artifact from github, so it has the HTTP archive download, and the git information

image

One issue with redirecting users to the HTTP url directly is that swift URLs downloads a file named just the UUID, with no extension, and the mime type was set to plain text. So my experience was that after downloading in chrome, I clicked on the file, and it opened in a text editor, but it worked fine opening with the archive manager.

### DIFF
--- a/sharing_portal/static/sharing_portal/artifacts.css
+++ b/sharing_portal/static/sharing_portal/artifacts.css
@@ -247,3 +247,7 @@
 .artifactVersion__doi {
     font-size: 12px;
 }
+pre code {
+    font-size: 12px;
+    white-space: inherit;
+}
--- a/sharing_portal/templates/sharing_portal/detail.html
+++ b/sharing_portal/templates/sharing_portal/detail.html
@@ -106,8 +106,12 @@
 
     {% for git_method in git_content %}
 <div>
-  <h4><i class="fa fa-code"></i>Checkout with git</h4>
+  <h4><i class="fa fa-code"></i>Download with git</h4>
+    <p class="artifactLaunchButton_details">
+    Clone the git repository for this artifact, and checkout the version's commit
+    </p>
     <pre><code>git clone {{git_method.remote}}
+# cd into the created directory
 git checkout {{git_method.ref}}</code></pre>
 </div>
     {% endfor %}
--- a/sharing_portal/templates/sharing_portal/detail.html
+++ b/sharing_portal/templates/sharing_portal/detail.html
@@ -97,6 +97,21 @@
       </p>
     {% endif %}
 
+    {% if http_content %}
+    <a class="artifactLaunchButton btn btn-primary" href="{{ download_url }}" target="_blank">Download Archive</a>
+    <p class="artifactLaunchButton_details">
+    Download an archive containing the files of this artifact.
+    </p>
+    {% endif %}
+
+    {% for git_method in git_content %}
+<div>
+  <h4><i class="fa fa-code"></i>Checkout with git</h4>
+    <pre><code>git clone {{git_method.remote}}
+git checkout {{git_method.ref}}</code></pre>
+</div>
+    {% endfor %}
+
     <div class="artifactVersions">
       <h4><i class="fa fa-files-o"></i> Versions</h4>
       <ol class="sidebarNav">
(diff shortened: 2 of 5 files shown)


super-cooper:
> One issue with redirecting users to the HTTP url directly is that swift URLs downloads a file named just the UUID, with no extension, and the mime type was set to plain text.

This is probably a result of Trovi uploading `application/octet-stream` rather than `application/tar+gz`. Apparently including the file extension in the name of the upload does not matter to Swift. I went ahead and fixed this in Trovi :D 

image

Unfortunately, archives that have already been uploaded using the API with this problem will not be fixed, I believe.

super-cooper:
> @super-cooper it's possible to change content-type for any Swift object w/o changing the object contents, see https://docs.openstack.org/api-ref/object-store/?expanded=create-or-update-object-metadata-detail

Oh neat! I imagine we only have a few archives with this problem. We could go in and fix them.

sharing_portal/views.py
```diff
@@ -488,16 +488,39 @@ def artifact(request, artifact, version_slug=None):
     )
     launch_url = preserve_sharing_key(launch_url, request)
 
+    # We use this download URL instead of the one from trovi so we can
```
super-cooper, line 491: This comment confused me. It looks like we _are_ using the download URL from Trovi, but we're just fetching it when the button is clicked rather than when the page loads (which is the right thing to do, imo).

super-cooper (approved):
Looks great :D

sharing_portal/static/sharing_portal/artifacts.css
```diff
@@ -247,3 +247,7 @@
 .artifactVersion__doi {
     font-size: 12px;
 }
+pre code {
+    font-size: 12px;
+    white-space: inherit;
```
diurnalist, line 252: A bit surprised this is needed, what existing styles are there on `<code>`?

sharing_portal/templates/sharing_portal/detail.html
```diff
@@ -97,6 +97,25 @@ <h4>Digital Object Identifier (DOI)</h4>
       </p>
     {% endif %}
 
+    {% if http_content %}
+    <a class="artifactLaunchButton btn btn-primary" href="{{ download_url }}" target="_blank">Download Archive</a>
```
diurnalist, line 101: I'm wondering if this would look less busy with `btn-small` (`btn-sm`?) and perhaps losing the `-primary` modifier so it's not green.

diurnalist (approved):
Generally looks good! The UI is starting to get a bit busy though over on the right rail. I think in the future we should explore a dropdown style mechanism, like GitHub uses on this page:

github-code-dropdown

diurnalist:
[Support direct download of artifact files w/o launching](https://trello.com/c/tv1P4AFs/2637-support-direct-download-of-artifact-files-w-o-launching)

sharing_portal/templates/sharing_portal/detail.html
Mark-Powers: `btn-sm` doesn't change anything. Everything in the sidebar is 100% width, so I'm not sure how this would look. Without `-primary` , there is no background color, and same with `-secondary`.

sharing_portal/views.py
Mark-Powers: I meant that this uses a portal download link e.g. `experiment/share/uuid/download` rather than directly setting the `href` to trovi's content URL.

sharing_portal/views.py
```diff
@@ -1089,3 +1112,19 @@ def create_artifact(request):
         "authors_formset": authors_formset,
     }
     return HttpResponse(template.render(context, request))
+
+
+@get_artifact
+@with_trovi_token
+def download(request, artifact, version_slug=None):
```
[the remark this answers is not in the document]
Mark-Powers: It doesn't require auth for public artifacts.

This does not increment a counter either at the moment, that is done manually in portal when a user is redirected to the `launch` URL.

Mark-Powers:
I agree the dropdown would be great in the future.

I'll add a card to update the trovi `content-type` setting.

sharing_portal/views.py
super-cooper: Yes, I understood that after I scrolled down in the PR and saw the download function. Upon initially reading the comment and seeing that we're doing a `reverse` instead of calling directly to Trovi in this function, I thought that we would not be fetching the link from Trovi at all for some reason. I did not realize we were just proxying the Trovi call.

PR #307, opened 2022-05-03
Add private icon next to a private artifact

Mark-Powers:
This shows an icon next to an artifact title when the artifact is private. The only other place this information is displayed is inside the sharing settings.

List view:
image

Detail view:
image

### DIFF
--- a/sharing_portal/static/sharing_portal/artifacts.css
+++ b/sharing_portal/static/sharing_portal/artifacts.css
@@ -72,6 +72,7 @@
 
 .cardItem__title {
     margin: 0 0 1rem;
+    display: flex;
 }
 
 .cardItem__body {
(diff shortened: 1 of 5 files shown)


super-cooper (approved):

super-cooper:
Does this icon have hovertext? I think that would be helpful.

sharing_portal/static/sharing_portal/artifacts.css
```diff
@@ -247,3 +248,15 @@
...
+
+.cardTitleTitle {
+    flex: 1;
+}
+.cardTitlePrivate {
+    flex: 1;
+    text-align: right;
+}
+.artifactDetailPrivate {
+    font-size: 32px;
```
diurnalist, line 260: What does it look like if we don't override this?

diurnalist:
We could also throw the icon into the "stats" component, which is re-used in both the list view and the detail view. That's where the icons are currently, I think it would be a decent place to put this one too.

sharing_portal/static/sharing_portal/artifacts.css
Mark-Powers: It's a bit too small. 
image

PR #311, opened 2022-05-16
Add check for allocation expiration in enforcement

Mark-Powers:
This prevents creating a lease that extends beyond someone's allocation.

### DIFF
--- a/balance_service/enforcement/exceptions.py
+++ b/balance_service/enforcement/exceptions.py
@@ -38,3 +38,9 @@ class EnforcementException(Exception):
 class BillingError(EnforcementException):
     msg_fmt = "Not authorized"
     code = 403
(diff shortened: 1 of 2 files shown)


diurnalist (approved):
Very clever! I'm glad the case was made for moving this code into portal rather than a separate service, it makes such things much simpler.

I don't think we surface this very well to end-users though, do we? Ideally this text is what would show up in the little red message box in Horizon, for example.

super-cooper (approved):
ezpz :)

Mark-Powers:
> I don't think we surface this very well to end-users though, do we? Ideally this text is what would show up in the little red message box in Horizon, for example

Correct, right now it just shows a generic error message.

PR #312, opened 2022-05-17
Add way to bulk add users on project page

Mark-Powers:
This adds a way to add multiple users at once to a project via a comma separated list.

A new button is next to the add user button.

image

When clicked, it opens a modal with a textfield.

image

Message summaries appear after form submission. All error messages are shown, which give detail of what users could not be added:

image

Or if there are no issues, that is displayed as well:
image

When adding a single user, nothing different is changed. The biggest concern with this change is that it could time out if too many emails are being sent, if users come across this we will need to defer sending mail to a celery task.

### DIFF
--- a/projects/forms.py
+++ b/projects/forms.py
@@ -269,3 +269,9 @@ class AddBibtexPublicationForm(forms.Form):
             return False
 
         return True
+
+class ProjectAddBulkUserForm(forms.Form):
+    username_csv = forms.CharField(
+        label="Usernames or emails",
+        widget=forms.Textarea(attrs={"placeholder": "Usernames seperated by commas"}),
+    )
--- a/projects/static/projects/js/bulk_invite.js
+++ b/projects/static/projects/js/bulk_invite.js
@@ -1,0 +1,8 @@
+'use strict';
+(function( window, $, undefined ) {
+  $('button[name="add_bulk_users_popup"]').on('click', function(e) {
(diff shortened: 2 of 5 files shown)


diurnalist (commented):
I like the error message, very descriptive 👍🏻 

How common would it be to paste a CSV formatted string? How would the user get that? Do you think newline-separated would be more user-friendly? What about supporting either?

zhenz (commented):
Should we just have one "Add users" button and remove adding a single user?

super-cooper (approved):

Mark-Powers:
@diurnalist
> How common would it be to paste a CSV formatted string? How would the user get that? Do you think newline-separated would be more user-friendly? What about supporting either?

I just looked at moodle and canvas, which I think are the two most popular course management systems. They both allow exporting as a spreadsheet, and from there you can easily copy one per line. So I think it is best to support one per line. To simplify things for now, we'll just support the one per line formatting, but if users request other formats, we can work on having it support both at once. I don't want to add extra complexity for this if it turns out only the one format is ever used.

@zhenz 
>Should we just have one "Add users" button and remove adding a single user?

My thought was if the existing input box is changed, it may confuse old users. And if the appearance does not change, then it isn't obvious we now support adding multiple users.

diurnalist (approved):
Cool! Thanks for checking what the conventions were on the export side :) :shipit:

PR #313, opened 2022-05-17
Handle missing openstack region config

diurnalist:
Local dev environments will not have this config, nor should they really.
However, not having this config means DB migrations cannot work (the
charge table migration will try to copy data from an external source.)

This will drop the region config from the dict if it's not defined on the
env, using the host field as the heuristic. Otherwise weird typeerrors
get raised when connecting to the DB.


super-cooper (approved):

Mark-Powers (approved):
Great change!

PR #314, opened 2022-05-17
Clean up hardware browser UI

diurnalist:
1. Replace human-friendly node type “summaries” with the actual node types in the big green box. Users are already confused about this enough, I think we can just stick to node type as a general filter mechanism.
2. Make the top line of buttons NOT green so they’re not semantically linked w/ the below buttons, which act as filters (the top row are all external links!)
3. Add the node type as a label in the details view so you can easily see which nodes are of which type.
4. Tightened up the headings a bit and renamed to match menu nav terminology.

### DIFF
--- a/g5k_discovery/templates/g5k_discovery/discovery.html
+++ b/g5k_discovery/templates/g5k_discovery/discovery.html
@@ -4,33 +4,37 @@
 {% block title %}Chameleon Resource Browser{% endblock %}
 
 {% block content %}
-<div>
-    <h1>Availability</h1>
-    <div class="row">
-        <div class="col col-sm-2">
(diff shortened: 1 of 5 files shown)


diurnalist:
Screen Shot 2022-05-17 at 4 30 04 PM
Screen Shot 2022-05-17 at 4 28 39 PM

super-cooper (approved):
A blessing! I've accidentally clicked on the Horizon buttons countless times :woozy_face:

Mark-Powers (approved):
Looks great! Does this also expose `node_name`? It looks like it in the code, but the screenshot makes it look like just `node_type`

diurnalist:
@Mark-Powers we do; if there is a node name it is shown instead of the uid

PR #316, opened 2022-05-24
Trovi usability

Mark-Powers:
This improves the search feature of trovi, and clears up the launch button missing for a logged out user.

### DIFF
--- a/sharing_portal/static/sharing_portal/index.js
+++ b/sharing_portal/static/sharing_portal/index.js
@@ -23,7 +23,7 @@
   function applyQueryFilter(filter) {
     const terms = filter.split(' ').filter(Boolean);
     document.querySelectorAll('.cardItem').forEach((el) => {
-      const matches = terms.some((term) => {
+      const matches = terms.every((term) => {
         return (term.length > 2) && el.dataset.search.includes(term);
       });
       el.classList.toggle('hidden', terms.length > 0 && !matches);
--- a/sharing_portal/static/sharing_portal/index.js
+++ b/sharing_portal/static/sharing_portal/index.js
@@ -21,10 +21,10 @@
   }
 
   function applyQueryFilter(filter) {
-    const terms = filter.split(' ').filter(Boolean);
+    const terms = filter.split(' ').filter((term) => term.length > 2).map((term) => term.toLowerCase());
     document.querySelectorAll('.cardItem').forEach((el) => {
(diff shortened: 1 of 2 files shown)


sharing_portal/views.py
```diff
@@ -482,7 +482,10 @@ def _parse_doi(artifact):
 @with_trovi_token
 @get_artifact
 def artifact(request, artifact, version_slug=None):
-    show_launch = request.user is None or has_active_allocations(request)
+    # Show the launch button if the user is logged out, or has active
+    # allocations. If the user is logged out, they will be asked to log in
+    # after clicking launch.
+    show_launch = not request.user.is_authenticated or has_active_allocations(request)
```
super-cooper, line 488: Is it no longer possible for the user to be `None` here?

super-cooper (approved):
LGTM! One question inline

sharing_portal/views.py
Mark-Powers: It seems that if the user is not logged in, then `request.user` is an [AnonymousUser](https://docs.djangoproject.com/en/4.0/ref/contrib/auth/#anonymoususer-object), which must be why this didn't work previously.

sharing_portal/static/sharing_portal/index.js
```diff
@@ -23,7 +23,7 @@
   function applyQueryFilter(filter) {
     const terms = filter.split(' ').filter(Boolean);
     document.querySelectorAll('.cardItem').forEach((el) => {
-      const matches = terms.some((term) => {
+      const matches = terms.every((term) => {
```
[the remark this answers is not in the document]
diurnalist: Side side note... `const` is used liberally here in order to enforce that variables are not re-used in the same lexical scope. Most JS engines these days will actually raise a TypeError if you try to reassign to a const. Linters can also catch this for you when authoring the code. It is one way to turn a potential runtime error (smashing a variable you didn't realize was in-scope, or perhaps not understanding scoping rules in JS) into a "compile-time" error, with the caveat that JS has no true compile phase, but at least it's when you're writing the code as opposed to when it's executing.

Mark-Powers:
@diurnalist Good point about filtering out small terms, and the lowercase was needed.

I added author names to the search term, which I think would be helpful. My concern though is that some names may confuse people (e.g. search for "power" would include artifacts where I am an author).

PR #317, opened 2022-05-26
Add a series of nested questions to the ticket create page

Mark-Powers:
With the goal of not annoying users or making it harder to submit
tickets (since the help desk is a frequently complimented feature),
this adds a series of questions at the top of the ticket create page
that can help a user solve their own issue.

Ideally, this text will help us by reducing ticket load, and help users
by quickly solving their problems. It also provides us a place to leave
feedback for what information should be included for certain types of
questions.

Here is what the UI looks like when the page loads:
image

The user can click each underlined bit of text to open the details. There are also buttons to expand/collapse all. Here is the UI with all items expanded.
image

Currently this is generated in JS to avoid having to manage a lot of nested tags by hand. The content is written in the JS file directly. Alternatively, we could set up a model for the data, and manage it via the admin panel, but that would add a lot to this change.

The initial content here is generated from past tickets where things could be user. I envision this as something we add to over time after seeing repeat tickets.


djangoRT/templates/djangoRT/ticketCreate.html
```diff
@@ -1,12 +1,37 @@
...
 {% include "djangoRT/navigation.html" %}
+<h1>Look up issue</h1>
+<div id="tree-container">
+  <p>
+    You can browse common issues here. Select each answer that fits your
+    problem. If nothing seems to fit, please submit a ticket.
+  </p>
+  <div id="expand_all" class="btn btn-sm btn-primary">Expand All</div>
+  <div id="collapse_all" class="btn btn-sm btn-primary">Collapse All</div>
+  <div id="tree"></div>
```
diurnalist, line 21: If you're worried about doing things by hand:

```html
{% for group in question_group %}
<details>
    <summary>{{ group.label }}</summary>
    {% if group.desc %}<p>{{ group.desc }}</p>{% endif %}
    {% for question in group.questions %}
    <details>
        <summary>{{ question.label }}</summary>
        <p>{{ question.desc }}</p>
    </details>
    {% endfor %}
</details>
{% endfor %}
```

I think this is arguably better, as it's pretty succinct, and JS could break more easily and doesn't seem to add much value for this UX. You would just need to ensure the 'desc' is treated as a "safe" string by Django so it doesn't escape the link elements.

djangoRT/static/djangoRT/css/djangoRT.css
```diff
@@ -59,3 +59,17 @@
 	content: "*";
 	color: red;
 }
+
+details {
+    margin-left: 2em;
+    display: list-item;
```
diurnalist, line 65: I think the built-in display, which renders a disclosure triangle, communicates better the functionality.

djangoRT/static/djangoRT/css/djangoRT.css
```diff
@@ -59,3 +59,17 @@
...
+}
+
+summary {
+    cursor: pointer;
```
diurnalist, line 69: This is good :+1:, I felt this should be a default in the built-in styling 🤷🏻

diurnalist (approved):
Looks like a nice improvement! Good ideas here.

I am not sure the expand/collapse buttons are helping much, they add a bit of clutter and it's not clear to what they refer (the text doesn't scream "I can expand!"). I'd recommend dropping them personally for simplicity.

djangoRT/templates/djangoRT/ticketCreate.html
Mark-Powers: I fully agree, I initially was working on a prototype that used more JS, and so it made sense there.

Mark-Powers:
@diurnalist My reasoning for the expand all is to make it easier to ctrl-f in case a user wants to search for an error message. Perhaps this isn't needed since it is nested. I agree the UI though doesn't fit perfectly, they may look better floated to the right.

djangoRT/templates/djangoRT/ticketCreate.html
Mark-Powers: The JS solution also allows for deeper nesting via recursion, but probably is overkill.

djangoRT/static/djangoRT/css/djangoRT.css
diurnalist: Ah yes, I missed this was a `display` attribute. Makes sense. So maybe it's `list-style`.  You could try `list-style-type: revert;` or `list-style-type: unset;` to try to get rid of any overrides being applied elsewhere? Otherwise I think this may work:

```
summary {
  display: list-item;
  list-style: disclosure-closed;
}
details[open] > summary {
  list-style: disclosure-open;
}
```

super-cooper (approved):
I think this change in and of itself is good. What I was imagining was to simply add some additional optional fields to the from like

Site
Lease ID
Instance ID
Network ID
What do you expect to happen
What actually happens
etc.

Just to let users know that that information is valuable and encourage them to include it if it's relevant to their issue. I think the issue with big text drop downs is that people may not read it. It will also need to be updated any time support policy/scope changes, or when we add new services.

I do think that the information included is very helpful as sort of an FAQ, though, and may prevent superfluous tickets, so that's :+1: :+1: :+1:

Mark-Powers:
I added some more form fields as @super-cooper mentioned, and updated the text field text, since it was just asking for the same info that is now in the new fields. 

Here is the updated summary triangle decoration (thanks @diurnalist for the suggestion/help):
image

super-cooper (approved):

PR #318, opened 2022-06-03
Add our domains to image csp

Mark-Powers:
Our images are uploaded either while on chameleoncloud.org
or www.chameleoncloud.org. Since the absolute URL is used in the CMS,
the images won't load when using the site via the other domain, since
the CSP only contained 'self' (along with some google domains). This
fix adds both domains, meaning that it should always be able to load the
images.

### DIFF
--- a/chameleon/settings.py
+++ b/chameleon/settings.py
@@ -867,6 +867,8 @@ CSP_FRAME_SRC = [
 ]
 CSP_IMG_SRC = [
     "'self'",
+    "https://chameleoncloud.org",
+    "https://www.chameleoncloud.org",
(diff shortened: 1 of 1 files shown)


diurnalist (commented):
The absolute URL is stored inside the CMS?? At upload time?? :O

diurnalist (approved):
Oh I see... it's about the embedded reference rather than the asset itself. Pretty lame. That means that if we ever changed domains no posts w/ images would be portable, which is quite a surprising consequence of (I assume) the design of django-filer.

super-cooper (approved):
:)

PR #319, opened 2022-06-03
Fix passing of ownership/artifact meta to jupyter

diurnalist:
Currently the workflow of creating a new version from an artifact
is busted if you are launching it for the first time and you own it.
Probably this only affects artifacts made before the migration, as
those are the ones that effectively had their local DB state cleared.
Though, this also would affect users who try to launch their own
artifact at the latest version, after publishing that version (because
it creates a new server, w/ a new notebook directory from scratch.)

Requires chameleoncloud/jupyterlab-chameleon#19 and
chameleoncloud/jupyterhub-chameleon#6


Mark-Powers (approved):
Looks simple enough

super-cooper (approved):

PR #322, opened 2022-06-17
Change the webinar link on user dashboard to link to the same page as the CMS menu

super-cooper:
I was unable to find a reverse link for the webinar page from the CMS menu. It looks like the page previously linked-to was used for webinar registration, and we use models from this app on the dashboard. Is this something we use at all anymore? Should it be purged from the user portal?

### DIFF
--- a/chameleon_cms_integrations/cms_menus.py
+++ b/chameleon_cms_integrations/cms_menus.py
@@ -100,7 +100,10 @@ class UserMenu(CMSAttachMenu):
 
         menu_id += 1
         n = NavigationNode(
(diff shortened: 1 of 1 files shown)


Mark-Powers (approved):
This change looks good! I don't think there is a reverse URL since the webinar page is CMS. There are some things in portal we should eventually remove: this and the old trovi DB models are two things that come to mind.

zhenz (approved):

super-cooper:
I created [a trello task](https://trello.com/c/2MI9mhq9/437-remove-legacy-models-and-views-from-portal) to remove the legacy webinar system, along with the legacy Trovi artifacts.

PR #323, opened 2022-06-18
send notification when user adds publications

zhenz:
We'd like to manually review user publications as soon as they are
entered. So after new user publications are successfully imported
to the portal DB, a notification will be sent to an internal
mailing list. We also make creating multiple publications an atomic
transaction to prevent duplications.


Mark-Powers (approved):
Looks good!

PR #324, opened 2022-06-21
Fix bug creating daypass project

Mark-Powers:
The relevant code got moved to the except block, and so would not run
correctly.


zhenz (approved):
LGTM!

PR #330, opened 2022-07-26
Trovi error messages are not very helpful in Portal

super-cooper:
Any unhandled Trovi exceptions in Portal would previously give an opaque error message to the user, which would not inform them of what actually went wrong. Now, any unhandled Trovi errors will have the API error message forwarded to the user.

`TroviException` is been modified to hold on to more error context.

The `handle_trovi_errors` decorator is implemented to wrap views with a `TroviException` handler, so that any `TroviExceptions` which would otherwise be unhandled are caught, and the API error message is reported to the user. This does not impede any explicit error handling within view functions.

There was some error reporting inside a generator expression that was never actually executed, so errors were never displayed.

The `PATCH` request generated by the sharing form also displayed an opaque error message, so handling has been referred to the decorator. This is the easiest way to handle patch errors, because it will explicitly display the fields which are incorrect.

### DIFF
--- a/sharing_portal/trovi.py
+++ b/sharing_portal/trovi.py
@@ -15,10 +15,22 @@ LOG = logging.getLogger(__name__)
 
 
 class TroviException(Exception):
-    def __init__(self, code, message):
-        self.code = code
-        self.message = message
+    def __init__(self, method, path, expected_code, actual_code, detail):
+        self.method = method
+        self.path = path
+        self.expected_code = expected_code
+        self.actual_code = actual_code
+        self.detail = detail
         super().__init__(self.message)
+
+    @property
+    def message(self):
+        return (f"{self.method} {self.path} {self.actual_code} returned, "
+                f"expected {self.expected_code}: {self.detail}")
+
+    @property
+    def code(self):
+        return self.actual_code
 
 
 def url_with_token(path, token, query=None):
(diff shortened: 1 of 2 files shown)


super-cooper:
[Trovi error messages are not very helpful in Portal](https://trello.com/c/jDPJv41S/2724-trovi-error-messages-are-not-very-helpful-in-portal)

sharing_portal/trovi.py
```diff
@@ -295,22 +307,26 @@ def delete_version(token, trovi_artifact_uuid, slug):
...
+                f"/artifacts/{artifact_id}/versions/{version_slug}/metrics/",
+                admin_token,
+                query=query,
+            )
         )
-    )
-    check_status(res, requests.codes.no_content)
+        check_status(res, requests.codes.no_content)
+    except TroviException as e:
+        # Allow user to launch artifact as to not interrupt, but log error anyway
```
Mark-Powers, line 328: good idea

Mark-Powers (approved):
Looks awesome! Does this display the path and status code to the user as well in portal? That may be something we want to hide from them.

super-cooper:
> Does this display the path and status code to the user as well in portal?

The handler does not. It should be logged in exceptions, but I don't see anywhere that it is displayed to the user.

PR #331, opened 2022-07-28
fix balance service check_alloc_expiration_date

zhenz:
Balance service doesn't allow user to create/extend lease beyond
active allocation expiration date - even though they have approved
allocation afterward. We fixed the issue by checking the
consecutive approved allocation.


Mark-Powers (approved):
nice fix!

super-cooper (approved):

PR #332, opened 2022-08-01
Allow artifact non-owners to review daypass requests

super-cooper:
Since the reviewer is really just reviewing a small allocation request, it is acceptable for them to review daypass requests for artifacts that they don't own. In review functions, we will fetch the artifact with an admin token to ensure that it is visible to portal. The only information that is exposed to the reviewer is the artifact title. This may be subject to change in the future

🏕 Also fixed a couple error messages that were not being output because they were stuck inside generator statements that are never executed.

### DIFF
--- a/sharing_portal/views.py
+++ b/sharing_portal/views.py
@@ -682,7 +682,8 @@ def request_daypass(request, artifact, **kwargs):
             )
         else:
             if form.errors:
-                (messages.add_message(request, messages.ERROR, e) for e in form.errors)
+                for e in form.errors:
+                    messages.add_message(request, messages.ERROR, e)
             return HttpResponseRedirect(
                 preserve_sharing_key(
                     reverse("sharing_portal:request_daypass", args=[artifact["uuid"]]),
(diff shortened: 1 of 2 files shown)


super-cooper:
[Daypass - Sharing key is not handed off to Trovi](https://trello.com/c/pMxe1Jpa/2708-daypass-sharing-key-is-not-handed-off-to-trovi)

sharing_portal/views.py
```diff
@@ -682,7 +682,8 @@ def request_daypass(request, artifact, **kwargs):
             )
         else:
             if form.errors:
-                (messages.add_message(request, messages.ERROR, e) for e in form.errors)
+                for e in form.errors:
+                    messages.add_message(request, messages.ERROR, e)
```
Mark-Powers, line 686: Do you know why the code doesn't work with the generator? Either way, I prefer your change, since iirc it makes the stack a log nicer in the even of an exception

Mark-Powers (approved):
Looks good! 

How do you feel about adding a line to [this paragraph](https://github.com/ChameleonCloud/portal/blob/master/sharing_portal/templates/sharing_portal/share.html#L67-L76), just stating "If you artifact is not public, the PI will be able to see the title for any reproducibility requests". That protects us in the event someone cares about keeping their artifact metadata private.

sharing_portal/views.py
super-cooper: Yes! Python generator expressions are static objects, which are not executed unless iterated through. So the previous code essentially created a generator object that would then promptly be thrown out by garbage collection without getting executed. Here is a demo:

```python3
>>> def side_effect(x):
...     print(x)
...     return x + 1
... 
>>> generator = (side_effect(i) for i in range(3))
>>> # The subject of the generator expression is never executed, because the generator is just being saved as an object
>>> data = list(generator)
0
1
2
```

super-cooper:
Added a data leak warning for private artifacts
image

Mark-Powers (approved):

PR #334, opened 2022-08-24
Add project tags

super-cooper:
This change adds project tags for portal projects, as described in the design doc. Tags replace the old project `Type` model. 

This change does not remove `Type`s, but does remove all references to them. The `Type`s need to stay for the migration to re-label old projects. They will need to be removed in a future change.

This change _does_ remove the field of science from projects, and all references to it. 

Some unused TAS code was removed, in order to make the change easier. 

Some TAS code has been renamed to be more generic, in order to make it more obvious what its function is. A lot of the TAS code was not able to be removed without significantly refactoring the entire project/allocation system. That should happen in a future change. 

The `Tag` model is basically the same as `Type`, but with an added `description` field. This is so we can use the shorthand where desired, or optionally include the long description to provide more context to users.

This change could have been smaller, but in the interest of not adding any more tech debt to portal since we're trying to cut it down... it is what it is.


super-cooper:
[Portal - Project Categorization](https://trello.com/c/g8jG2Edv/2751-portal-project-categorization)

Mark-Powers (approved):
Nice work! Great that so much is being deleted. Looks like this is mostly a renaming. My only comment really is regarding the using of the term JSON. I think JSON would be more appropriate for a JSON string or something, here we are just storing data in dicts, not really related at all to JSON. Can you also summarize again why it uses the object representation and the dict one again? It's been a while since I dived into the code.

super-cooper:
> Can you also summarize again why it uses the object representation and the dict one again? It's been a while since I dived into the code.

It's a legacy artifact from when it had to go back and forth between TAS. It should have been converted at the last possible second and used as a Django model everywhere in Portal imo, but so much of the project/allocation code expects a dict, and the rest of it expects a model. I originally did try to remove that code, but it would have required me to refactor most of the project-allocation code. A lot of behavior is built into the mapper classes, which I'm pretty sure technically don't need to exist anymore. That behavior would just have to be factored out. It seemed out of scope of this change, so I just decided to minimize the tech debt where I could so that we could remove the remnants of TAS at a later time. It will be in my list of tech debt for the portal systems review.

super-cooper:
> Could you make sure we back up the projects, field of science, and the old "type" tables before deploying to prod?

Yes, I will be sure to back everything up before deploying :)

PR #338, opened 2022-09-06
Add task to automatically tag 

super-cooper:
This patch adds a task to automatically tag any projects tagged as "Other." It will tag the project, mark it as `automatically_tagged`, and then send an email to the project's PI to inform them that their project has been tagged, and tell them where they can go to change it. Each PI will only receive one email. 

If the PI has **one** project, they will receive an email with a subject line like this:

image

With content like:

image

If the PI has **multiple** projects, they will receive an email with a subject line like this:

image

With content like: 

image

## How it works
The tagging is done with [nltk](https://www.nltk.org). No machine learning is used, since obtaining training data would be more difficult than just manually tagging the project. The task uses nltk's word normalizing utilities to reduce keywords to their root, and then count matches against a list of keyword roots for each tag. The tag with the most matches is chosen, and the project is marked as automatically tagged.

### Known issues
Against actual project descriptions, this method is quite accurate, but not perfect. To counteract this, we send the emails to the PIs, and we are also waiting until the 16th to run this task. The `automatically_tagged` attribute is added in case we ever want to filter out any projects that were not tagged by hand, in order to get more accurate metrics or just tag them ourselves.


super-cooper:
[Portal - Project Categorization Migration](https://trello.com/c/ZI1P8pOm/2770-portal-project-categorization-migration)

projects/migrations/0013_add_tag.py
```diff
@@ -141,6 +141,8 @@ def create_tags(apps, _):
 
     tag_model.objects.bulk_create(tags)
 
+    if type_model.objects.count() == 0:
```
Mark-Powers, line 144: Why is the migration being updated? This already has been run on prod right?

projects/tasks.py
```diff
@@ -127,3 +130,587 @@ def handle_too_many_daypass_users(artifact_uuid):
...
+    
+    <p>
+    If our selections do not seem correct to you, you may freely update your tags by
+    visiting the links to your projects above.
+    </p>
+    """
+    )
+
+
+def automatically_tag_projects():
```
Mark-Powers, line 206: To run this manually, you should be able to `exec` into the django shell, and them `from project.tasks import automatically_tag_projects; automatically_tag_projects()`.

Mark-Powers (approved):
Looks good! I left a note on running this.

projects/migrations/0013_add_tag.py
super-cooper: Yeah, it won't run again. This is just to catch if people want to run on their local deployments. The migration fails if you have an empty types/projects table. This just short circuits it.

zhenz (approved):
LGTM! I like the idea of adding the "automatically_tagged" field :) 

Do we have a "dry run" mode for this task? Or did you already test against prod db? Just want to make sure we don't send incorrect emails to users.

super-cooper:
> Do we have a "dry run" mode for this task? Or did you already test against prod db? Just want to make sure we don't send incorrect emails to users.

No, but that's an excellent idea, so I've just added a dry mode flag which will not save the models or send emails.

PR #339, opened 2022-09-12
[trovi] add unique users metric

zhenz:


super-cooper (approved):
I really love the new icons!! I also love that there are tool tips now!!!!!

But I wonder if we can have a more descriptive tool tip than "users"? I can't think of one though.

sharing_portal/templates/sharing_portal/includes/stats.html
```diff
@@ -2,18 +2,25 @@
 
 <div class="artifactStats">
   <div class="artifactStats__meta">
-    <span><i class="fa fa-repeat"></i>
+    <span title="launches"><i class="fa fa-rocket"></i>
```
Mark-Powers, line 5: We should use the bootstrap `tooltip` field. Here is the example of how we use it for the private icon. This way it'll have a nicer looking popup.

```
<i class="fa fa-eye-slash artifactDetailPrivate" data-toggle="tooltip" data-placement="top" title="This artifact is only visible for those chosen by the author"></i>
```

Mark-Powers (changes requested):
See the comment about adding the tooltip info so it matches the rest of the site.

I also think we should use phrases for each of these, instead of single words, maybe "Total launches", "Total users" or "Unique users", and then "Number of versions"?

Mark-Powers (approved):

PR #340, opened 2022-09-15
Fix ordering of versions in sharing page

super-cooper:


super-cooper:
[Artifact versions are listed in reverse order on share page](https://trello.com/c/w2vPIcQk/2791-artifact-versions-are-listed-in-reverse-order-on-share-page)

Mark-Powers (commented):
`_artifact_display_versions` is used only in the share/edit screen? Just to clarify, after this change, the main version selector on the detail page will list from oldest to newest, but in the share screen, the newest is first?

Can you update the docstring for this function to clarify?

Also, can

Mark-Powers:
Yes, I see the trello description, my concern is there are 2 ways versions are listed, one of those should be reversed. This change removes the `reversed` call from this view, and I'm unsure if there is a reversal someone else for the other view.

super-cooper:
? Why should one be reversed?

super-cooper:
This function is only used for the sharing page and nowhere else. The versions are listed newest-to-oldest on the detail page, and this change makes the sharing page consistent with that ordering.

Mark-Powers (approved):
OK, newest to oldest everywhere sounds fine to me. Historically, both were not that way, but I think it's fine and doesn't break any documentation or anything.

super-cooper:
I think they actually were ordered the same way before. Right now with the current ordering, there is a bug that prevents you from creating a DOI for the newest version if there's already a DOI for an older version. This patch fixes that.

PR #341, opened 2022-09-15
Fix DOI display in Artifact detail page

super-cooper:
The artifact detail page was not properly displaying DOIs. This patch fixes that.

It also alters the previous behavior. The detail page used to show the DOI for whichever version was the most recent, regardless of which version was selected. It now shows the DOI for the selected version only.

The DOI check for making an artifact private on the share page was also removed. Artifacts with DOIs are allowed to be private, but versions with DOIs are still shown.

### DIFF
--- a/sharing_portal/views.py
+++ b/sharing_portal/views.py
@@ -194,9 +194,7 @@ def _compute_artifact_fields(artifact):
     artifact["is_chameleon_supported"] = any(
         label == "chameleon" for label in artifact["tags"]
     )
-    artifact["is_private"] = artifact["visibility"] == "private" and not _parse_doi(
-        artifact
-    )
+    artifact["is_private"] = artifact["visibility"] == "private"
     return artifact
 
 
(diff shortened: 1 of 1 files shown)


super-cooper:
[DOI display on artifact detail page is broken](https://trello.com/c/2pPkHS7M/2794-doi-display-on-artifact-detail-page-is-broken)

Mark-Powers (approved):
I think this change is OK, but I don't understand what the bug here was. It just is changing the functionality a bit, but I think this new functionality makes more sense.

The `is_private` change is fine too, as that's no longer something portal should be managing.

super-cooper:
> but I don't understand what the bug here was

It's a bug fix through behavior change! The bug was that this page was not displaying DOIs at all. Check out this artifact:

https://chameleoncloud.org/experiment/share/ae64bc9b-8315-41dc-96ec-abd86259ba3f

It has a DOI attached to its most recent version:

image

But it does not display this on the detail page.

PR #343, opened 2022-09-21
Fix tag widget in artifact edit form

super-cooper:
There was previously a bug with the tag widget, which would cause it to initially be empty. Because of this, saving the artifact without any changes would cause the tags to be removed from the artifact. Now, the widget will be initialized with whatever tags the artifact had previously.

The widget has also been changed to a multiple-choice checkbox. The multi-select menu thing seems more difficult to use, as it required users to CTRL+Click on multiple options in a tiny scrolling field. It was also not possible to view all of the tags at once in this view.

### DIFF
--- a/sharing_portal/forms.py
+++ b/sharing_portal/forms.py
@@ -1,5 +1,5 @@
 from django import forms
-from django.forms import widgets
+from django.forms import widgets, CheckboxSelectMultiple
 
 from projects.models import Project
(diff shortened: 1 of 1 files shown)


super-cooper:
[Artifact tags get reset on edit](https://trello.com/c/WnXXffnl/2792-artifact-tags-get-reset-on-edit)

Mark-Powers (approved):
Good change! I like the CheckboxSelect.

PR #344, opened 2022-09-21
Fix crash on artifact edit

super-cooper:
The error handling for artifact edits was not updated in line with the changes from ccee53b5813872a62b210256c2e765d2ce975cf9.


super-cooper:
[Portal artifact edit page always reports an error, even if the edit succeeds](https://trello.com/c/bvFMIJfA/2793-portal-artifact-edit-page-always-reports-an-error-even-if-the-edit-succeeds)

Mark-Powers (approved):
Good catch!

PR #351, opened 2022-10-17
publication updates

zhenz:
- add fields to publication model
- add script for updating citation numbers
- add scripts to import user publications from Scopus and Semantic Scholar


projects/user_publication/utils.py
```diff
@@ -0,0 +1,108 @@
...
+                SELECT u.first_name, u.last_name, p.id AS project_id, p.charge_code
+                FROM auth_user AS u
+                JOIN projects_project AS p
+                ON u.id = p.pi_id
+            ) AS proj
+            JOIN allocations_allocation AS a
+            ON proj.project_id = a.project_id
+            GROUP BY proj.project_id
+        """
+        cursor.execute(sql)
```
super-cooper, line 41: I feel very strongly that we should avoid this. As we discussed during the systems review, raw SQL code is a very easy vector for bugs. We should use the Django ORM since we already have models for these tables.

I don't know SQL very well, so I'm not really 100% positive what this whole function does. Does it just group projects by PI?

If so, we could do something like

```python3
pi_projects = defaultdict(list)
for project in Project.objects.all():
    pi = project.pi
    key = f"{pi.first_name.lower()}.{pi.last_name.lower()}"
    pi_projects[key].append(project)
```

projects/user_publication/utils.py
```diff
@@ -0,0 +1,108 @@
...
+            if len(last_name) == 1:
+                last_name = last_name[0]
+            else:
+                last_name = last_name[1]
+
+            # Vijay uses a different name for his publications
+            if last_name.lower() == "pillai" and r["first_name"].lower().startswith(
+                "vijay"
+            ):
+                last_name = "chidambaram"
```
super-cooper, line 62: This seems unsustainable if more users do something like this. Is there any way to do this without hardcoding?

super-cooper (dismissed):
Thank you for all of this! Looks great for the most part, but there are some concerns that should be addressed.

projects/user_publication/citation.py
```diff
@@ -0,0 +1,117 @@
...
+                        f"from {pub.semantic_scholar_citations} "
+                        f"to {citation_cnt}"
+                    )
+                )
+            else:
+                pub.semantic_scholar_citations = citation_cnt
+                pub.save()
+
+
+def update_citation_numbers(dry_run=True):
```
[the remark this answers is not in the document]
zhenz: I have instructions here: https://docs.google.com/document/d/1tWxNbhg78u2m1ZvC0_yYILRDNitMh9_zwnCL2-bukEU/edit#bookmark=id.gj3ydawxzbl7.

projects/user_publication/utils.py
```diff
@@ -0,0 +1,108 @@
...
+    "application-based qos support with p4 and openflow: a demonstration using chameleon",
+    "application-based qoe support with p4 and openflow",
+    "overcast: running controlled experiments spanning research and commercial clouds",
+    "managing allocatable resources",
+    "a case for integrating experimental containers with notebooks",
+    "managing allocatable resources ( invited paper )",
+    "next generation clouds, the chameleon cloud testbed, and software defined networking (sdn)",
+    "chi-in-a-box: reducing operational costs of research testbeds",
+    "migrating towards single sign-on and federated identity",
+]
```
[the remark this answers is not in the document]
zhenz: sometimes, different academic search engines put the titles in different ways. for example, "managing allocatable resources" and "managing allocatable resources ( invited paper )".

projects/user_publication/semantic_scholar.py
```diff
@@ -0,0 +1,245 @@
...
+    "1c91f729e7f51cb6fb709dfedcddb9bde10d1914": "Chameleon: A Scalable Production Testbed for Computer Science Research",
+    "b99e844351487d620a69bf25a4df538b418b30f6": "Chameleon: A Large-Scale, Deeply Reconfigurable Testbed for Computer Science Research",
+    "6738ab2ba7ab2971153fd951e02af62e43b15e5b": "Next Generation Clouds, the Chameleon Cloud Testbed, and Software Defined Networking (SDN)",
+}
+
+
+def _parse_authors(author_names):
+    author_keys = []
+    for name in author_names:
+        n = name.rsplit(" ", 1)
```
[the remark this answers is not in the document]
zhenz: I think the simplest way is we grab the last word in a full name and say that is the surname. And I think that's true for most cases.

super-cooper:
Updates finally pushed. One thing I notice about Semantic Scholar is that most of the authors are in the format of "F. Lastname". So the matching against our user database doesn't really work. It finds about 5 publications and the rest are ignored. This could be something we update in the future, but then we would be matching against incomplete data, which could result in false positives.

projects/admin.py
```diff
@@ -27,15 +29,26 @@ class PublicationAdmin(ProjectFields, admin.ModelAdmin):
...
+        "approved_with",
+        "scopus_citations",
+        "semantic_scholar_citations",
     )
 
     ordering = ["project__charge_code", "-year", "-entry_created_date"]
     list_display = ("title", "project_charge_code", "year", "entry_created_date")
 
 
+class ChameleonPublicationAdmin(admin.ModelAdmin):
```
Mark-Powers, line 47: What exactly is this being used for? I see there is also `PublicationAdmin` and I don't really understand what `ChameleonPublicationAdmin` would be used for.

Mark-Powers: Ah I see these are the ones that Chameleon has published.

projects/user_publication/semantic_scholar.py
```diff
@@ -0,0 +1,245 @@
...
+            surname = n[1]
+            givenname = n[0][0] + "."
+        except Exception:
+            continue
+        author_keys.append((givenname.lower(), surname.lower()))
+    return author_keys
+
+
+def _search_semantic_scholar(query):
+    url = "http://api.semanticscholar.org/graph/v1/paper/search"
```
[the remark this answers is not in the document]
Mark-Powers: This still is unchanged btw

Mark-Powers (approved):
Looks great! Good work translating the raw SQL queries. The database models look very good.

projects/admin.py
super-cooper: Were you able to check out the admin views? Docker was not cooperating with me and I was not able to look at them locally.

super-cooper:
[Get on top of research impacts situation](https://trello.com/c/EScAsmd9/2819-get-on-top-of-research-impacts-situation)

PR #373, opened 2023-02-14
Report publications imported by semantic and scopus

AnishReddyRavula:
Made following changes
- Overall - Gets publications from each script, dedupe them, reports them, and saves them in DB
- Unicode issues with names and titles - Using unidecode library to decode them to ASCII chars to perform operations with SQL (LIKE)
- Modified each script to return the publication models instead of doing something with them in their individual scripts
- A calling method to get the models, de-duplicate them, report them (CSV and nicely print), and save the publications to DB


projects/user_publication/publications.py
```diff
@@ -0,0 +1,43 @@
...
+            logger.info(
+                f"removing duplicate publication {p} - original publication - {o_pub}"
+            )
+            pubs.remove(p)
+    return pubs
+
+
+def import_pubs(file_name, dry_run=True):
+    pubs = []
+    reports = []
```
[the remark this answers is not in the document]
AnishReddyRavula: will remove this in the next commit

projects/user_publication/publications.py
```diff
@@ -0,0 +1,43 @@
...
+    pubs.extend(scopus.pub_import(dry_run))
+    pubs.extend(semantic_scholar.pub_import(dry_run))
+    pubs = _get_unique_pubs(pubs)
+    report_publications(pubs)
+    export_publications(pubs, file_name)
```
Mark-Powers, line 39: Could this be toggled with a flag? I imagine we won't want it to run every time.

AnishReddyRavula: making param: file_name to be optional

projects/user_publication/semantic_scholar.py
```diff
@@ -131,8 +132,12 @@ def _get_pub_type(types, forum):
     return "other"
 
 
-def _save_publication(publication, dry_run=True):
+def _get_pub_model(publication, dry_run=True):
     title = publication.get("title")
+    decoded_title = unidecode(title)
```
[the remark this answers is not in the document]
AnishReddyRavula: I agree

projects/user_publication/semantic_scholar.py
```diff
@@ -131,8 +131,8 @@ def _get_pub_type(types, forum):
     return "other"
 
 
-def _save_publication(publication, dry_run=True):
-    title = publication.get("title")
+def _get_pub_model(publication, dry_run=True):
+    title = utils.decode_unicode_text(publication.get("title"))
```
super-cooper, line 135: Will this require us to re-encode the text every time we display it?

projects/user_publication/utils.py
```diff
@@ -3,12 +3,37 @@
 import re
 from collections import Counter
 from difflib import SequenceMatcher
+from unidecode import unidecode
```
super-cooper, line 6: It looks like we're not taking a dependency on this in our poetry config even though it's a third-party library.

projects/user_publication/utils.py
```diff
@@ -73,6 +99,27 @@ def guess_project_for_publication(authors, pub_year):
     return counter.most_common(1)[0][0]
 
 
+def report_publications(pubs, display=True):
+    line_format = "{0:18} : {1}\n"
+    for pub in pubs:
+        pd = pub.__dict__
+        report = [line_format.format(ck, pd.get(ck)) for ck in PUBLICATION_REPORT_KEYS]
```
super-cooper, line 106: For this, it might be a good idea to implement `Publication.__repr__` so we could reuse it for logging in other places if desired.

super-cooper (changes requested):
It looks like this change checks against duplicates during withing the import, but does not check against publications already in the database. I think this is a critical thing to do because it is possible that this script will cause the database to slowly accrue duplicates over time.

We could implement this in a simple way. Pseudocode:
```
for new_model in new import:
    query = Publications.objects.get(title__iexact=new_model.title, year=new_model.year)
    if not query.exists():
        new_model.save()
```
This should allow us to check for duplicates against the existing models and the new ones at the same time.

projects/user_publication/semantic_scholar.py
Mark-Powers: This is a good question. My understanding is we only need to decode before putting things into the database, as it's mysql with the unicode issues. It looks like this function is called with a publication that might be either from scopus/semantic scholar or also the database?

projects/user_publication/utils.py
Mark-Powers: Good catch, @AnishReddyRavula not sure if that file is just missing from the commit, or you'll need to run `poetry add --lock` to add it https://python-poetry.org/docs/cli/#add.

Mark-Powers (changes requested):
Looks good, but there are a few comments from myself and Adam to respond to. The dependency issue needs to be resolved too.

projects/user_publication/utils.py
AnishReddyRavula: For some reason, I can import this library without making any changes to the poetry.

Mark-Powers: It might be a dependency of something else we use (it wouldn't surprise me if django uses it), but the requirement should be made explicit in case whatever it's included with drops it.

projects/user_publication/semantic_scholar.py
AnishReddyRavula: Yes, this is only from scopus/semantic and not from the database. If we indent to show it in the website with same encoding then I think changing database to allow unicode chars is the way to go.

AnishReddyRavula:
> 



> It looks like this change checks against duplicates during withing the import, but does not check against publications already in the database. I think this is a critical thing to do because it is possible that this script will cause the database to slowly accrue duplicates over time.
> 
> We could implement this in a simple way. Pseudocode:
> 
> ```
> for new_model in new_import:
>     query = Publications.objects.filter(title__iexact=new_model.title, year=new_model.year)
>     if not query.exists():
>         new_model.save()
> ```
> 
> This should allow us to check for duplicates against the existing models and the new ones at the same time. It also allows us to take advantage of the speed of the database to do the check instead of doing it in python.

This is already being performed in [L158-semantic_scholar.py](https://github.com/ChameleonCloud/portal/blob/report_pubs/projects/user_publication/semantic_scholar.py#L158
)
[L98-scopy.py](https://github.com/ChameleonCloud/portal/blob/report_pubs/projects/user_publication/scopus.py#L98)
`Publication.objects.filter(title=title, project_id=proj).exists()`

projects/user_publication/utils.py
```diff
@@ -99,14 +99,11 @@ def guess_project_for_publication(authors, pub_year):
...
-def report_publications(pubs, display=True):
+def report_publications(pubs):
     line_format = "{0:18} : {1}\n"
     for pub in pubs:
-        pd = pub.__dict__
-        report = [line_format.format(ck, pd.get(ck)) for ck in PUBLICATION_REPORT_KEYS]
-        if display:
-            print(*report, sep="")
-            print("")
+        print(pub.__repr__)
```
super-cooper, line 105: You should be able to just `print(pub)` since you implemented `__repr__`

```python
>>> class Foo:
...     def __repr__(self):
...         return "bar"
... 
>>> print(Foo())
bar

Comment (AnishReddyRavula) [reply projects/user_publication/utils.py:105]: Because __str__ is also implemented, it prints out __str__() but not __repr__

File: projects/models.py [hunk 8]
```diff
@@ -267,6 +282,14 @@ class Publication(models.Model):
 
     def __str__(self) -> str:
         return f"{self.title}, {self.author}, In {self.forum}. {self.year}"
+
```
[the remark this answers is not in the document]
AnishReddyRavula: will remove in next branch

Mark-Powers (approved):

PR #375, opened 2023-02-20
Gscholar pubs

AnishReddyRavula:
Added script to get publications from gscholar
- Finds all the references of chameleon publications from gscholar
- Uses scraper API or free proxies to use scholarly package and get the results from google scholar
- Using a _handle_proxy_call wrapper before each API call so every call goes through proxy
- Moved few methods to utils so other files can use it.


projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+
+from projects.models import ChameleonPublication, Publication
+from projects.user_publication import utils
+
+logger = logging.getLogger("projects")
+
+MAX_RETRIES = 10
+
+
+def handle_scrapper():
```
Mark-Powers, line 17: the name of this method is a bit confusing, since it's main purpose is to get the API key. how about something like `get_scraper_api_key`?

Mark-Powers (changes requested):
a few minor comments, but overall nice work!

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+            except StopIteration:
+                logger.info(f"End of iteration. Got {len(citations)} / {no_citations}")
+                return citations
+
+    @_handle_proxy_reload
+    def _get_next_cite(self, cites_gen):
+        return next(cites_gen)
+
+    @_handle_proxy_reload
+    def _get_bib(self, pub):
```
super-cooper, line 125: This sounds like it should fetch the bibtex from the pub and return the bibtex, but it actually fetches the bibtex, adds it to the pub, and returns the pub.

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+        bibt = pub['bibtex']
+        bibt = utils.decode_unicode_text(bibt)
+        match = re.search(r'author\s*=\s*\{([\w\s-,]+)\}', bibt)
+        if match:
+            authors_t = match.groups()[0]
+        else:
+            # this mostly never happens - if it does - log it
+            logger.info(f"cannot parse bibtex, so ignoring - {bibt}")
+            return
+        if "and" in authors_t:
```
super-cooper, line 148: You don't need to check for this. Running `authors_t.split(" and" )` will return the same result as what you've done in your else statement if `"and"` does not appear in the string. Also, what does the `t` in `authors_t` mean? There's no need to shorten variable names so much. It's better to be clear and unambiguous for future readers.

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+            authors = authors_t.split(" and ")
+        else:
+            authors = [authors_t]
+        parsed_authors = []
+        for author in authors:
+            parsed_authors.append(utils.parse_author(author))
+        return authors
```
super-cooper, line 155:
```python
parsed_authors = [utils.parse(author) for author in authors]
```

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+            added_by_username="admin",
+            forum=pub["bib"]["booktitle"],
+            link=pub["pub_url"],
+            source="gscholar",
+            status=Publication.STATUS_IMPORTED,
+        )
+
+
+def pub_import(
+    dry_run=True, year_low=2014, year_high=datetime.now().year
```
super-cooper, line 174: Same thing here as above regarding default arguments

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+    - Checks if there is a project associated with the publication
+
+    Args:
+        dry_run (bool, optional): Defaults to True.
+        year_low (int, optional): Defaults to 2014.
+        year_high (int, optional): Defaults to datetime.now().year.
+    """
+    gscholar = GoogleScholarHandler()
+    pubs = []
+    for chameleon_pub in ChameleonPublication.objects.filter(id=14):
```
super-cooper, line 187: This filters for the one Chameleon publication that has ID `14`.

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+                continue
+            proj = utils.guess_project_for_publication(
+                authors, cited_pub["bib"]["pub_year"]
+            )
+            if not proj:
+                continue
+            if ChameleonPublication.objects.filter(title__iexact=cited_pub_title).exists():
+                logger.info(f"{cited_pub_title} is a chameleon publication - ignoring")
+                continue
+            if Publication.objects.filter(title=cited_pub_title, project_id=proj).exists():
```
super-cooper, line 203: Same thing here re: Mark's comment about what constitutes a duplicate.

super-cooper (changes requested):
Mostly looks good with lots of little nits regarding naming and clarity.

projects/user_publication/gscholar.py
AnishReddyRavula: Right. This was used for debugging

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+    if scraper_api:
+        with scraper_config.open('w') as s_config:
+            s_config.write(scraper_api)
+    return scraper_api
+
+
+class GoogleScholarHandler(object):
+    def __init__(self):
+        """To make google scholar calls using proxy"""
+        self.pg = ProxyGenerator()
```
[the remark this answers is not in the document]
AnishReddyRavula: Using ```self.proxy```

projects/user_publication/gscholar.py
```diff
@@ -0,0 +1,207 @@
...
+        else:
+            self.pg.FreeProxies()
+            logger.info("Using Free proxy")
+        self.scholarly.use_proxy(self.pg)
+        self.retries = 0
+        self.citations = []
+
+    def _handle_proxy_reload(func):
+        def inner_f(self, *args, **kwargs):
+            while self.retries < MAX_RETRIES:
```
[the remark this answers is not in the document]
AnishReddyRavula: If it exceeds maximum retries shouldn't it be exiting raising an error?

projects/user_publication/gscholar.py
AnishReddyRavula: Right. it means ```authors_text``` will change it to ```authors_match```

projects/user_publication/gscholar.py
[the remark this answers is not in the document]
super-cooper: Nothing seems to guarantee that the exception is thrown. This while loop seems to count to 10 and then always return a symbol called `resp`. What happens if it counts all the way to 10? If resp doesn't exist, it will throw a `NameError` which is not a clear explanation of what went wrong. I believe this code is also meant to be run manually, so even if it throws, I think the object owning this method will still exist, no? If I try to run the same code again, it will never work because retries has not been reset to 0.

projects/user_publication/publications.py
```diff
@@ -1,43 +1,44 @@
...
+import logging
+from projects.user_publication import scopus, semantic_scholar, gscholar
+from projects.user_publication.utils import export_publications, report_publications
+
+logger = logging.getLogger(__name__)
+
+
+def _is_same_publication(o_pub, d_pub):
+    # check if original publication is same as duplicate
+    if o_pub.title == d_pub.title and o_pub.year == d_pub.year:
```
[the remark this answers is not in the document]
AnishReddyRavula: I will talk to Kate and we can decide on a good strategy for this.
Will include that change in the next PR - google citations branch

projects/user_publication/gscholar.py
AnishReddyRavula: as included in Mark's comment I will talk to Kate and we can decide on a good strategy for this.
Will include that change in the next PR - google citations branch

Mark-Powers (approved):

super-cooper (approved):

PR #380, opened 2023-03-07
Citation counts

AnishReddyRavula:
- Changes in Publication models to hold a boolean for each source if publication is available in that source
- Added 2 other choices in 'status' field - Duplicate and Rejected

- Change in Google scholar - authors from gscholar bib has unicode charachters in ASCII with escape sequence "\" - I was not able to get the example right now but its in \u format but somethig like \{c}
- Using unicodedata to normalize and decode to ascii

- Added Citation count functionality in GoogleScholar class
- Added retries for SemanticScholar

Todo: DeDuplication system needs to be in place


projects/user_publication/gscholar.py
```diff
@@ -120,33 +124,83 @@ def get_authors(self, pub):
...
-            bibtex_source="{}",
+            publication_type=pub_type,
+            bibtex_source=pub['bib'],
             added_by_username="admin",
-            forum=pub["bib"]["booktitle"],
-            link=pub["pub_url"],
-            source="gscholar",
+            forum=forum,
+            link=pub.get("pub_url", ''),
+            source=Publication.G_SCHOLAR,
```
Mark-Powers, line 160: What exactly is source for? I thought publications can have multiple sources.

projects/user_publication/gscholar.py
```diff
@@ -164,28 +218,30 @@ def pub_import(
...
                 logger.info(f"{cited_pub_title} is a chameleon publication - ignoring")
                 continue
             # this is to find if there is a publication already in database matching to
             # this title and project however this needs to be changed after looking
             # at some examples and how to better handle these dumplicates
-            if Publication.objects.filter(title=cited_pub_title, project_id=proj).exists():
-                logger.info(f"{cited_pub_title} is already in Publications table - This check might be outdated.")
+            pub_exists = Publication.objects.filter(title=cited_pub_title, project_id=project)
+            if pub_exists:
+                add_to_all_sources(pub_exists[0], Publication.G_SCHOLAR)
```
Mark-Powers, line 241: why is index 0 used here? I think the name `pub_exists` also should be `existing_publication` to make it more clear.

Mark-Powers (changes requested):
In general, I think we should avoid using setattr. I'm not sure `setattr(pub, G_SCHOLAR, True)` is better than `pub.google_scholar = True` and similar. I thought we had discussed turning sources into a many-to-one relation, where there is a new table that references publication storing the publication id and source type? That would avoid the need to change the Publication model for each new source.

projects/user_publication/gscholar.py
AnishReddyRavula: Where the publication first sources from -user_reported etc.,

projects/user_publication/gscholar.py
AnishReddyRavula: because it is a filter, it will return a <type:queryset> so the first element is needed. Because there can be duplicates in the table now - using a `get()` will result in error

projects/user_publication/utils.py
```diff
@@ -26,10 +27,20 @@
 ]
 
 
+def add_to_all_sources(pub, source):
```
super-cooper, line 30: I'm not a fan of abstractions like this. It would be much simpler and more robust to just set the field and save it rather than using `setattr`.

projects/user_publication/utils.py
```diff
@@ -42,7 +53,11 @@ def guess_project_for_publication(authors, pub_year):
     in the publication's list of authors.
     """
     from projects.models import Project, ProjectPIAlias
-
+    try:
+        pub_year = int(pub_year)
```
super-cooper, line 57: Under what circumstances would this happen? I think if the caller passes an invalid year, I would consider that an invariant, and the code should not continue executing.

Are there any valid scenarios where a publication could have a non-integer year?

projects/user_publication/gscholar.py
```diff
@@ -120,33 +124,83 @@ def get_authors(self, pub):
...
+        if 'ENTRYTYPE' in pub['bib']:
+            entry_type.append(pub['bib']['ENTRYTPE'])
+        if 'pub_type' in pub['bib']:
+            entry_type.append(pub['bib']['pub_type'])
+        forum = ''
+        if 'booktitle' in pub['bib']:
+            forum = pub['bib']['booktitle']
+        if 'journal' in pub['bib']:
+            forum = pub['bib']['journal']
+        pub_type = utils.get_pub_type(entry_type, forum)
```
super-cooper, line 149: I didn't notice this in the previous code review, but there are now two functions called `get_pub_type` that both do almost the exact same thing in the publication utils module. This needs to be cleaned up. The functions either need to be merged or their names need to be changed. They should also be in the same scope (i.e. in the `PublicationUtils` class, since that's where the existing function is).

super-cooper (changes requested):
At a high level, the changes are good, but the code requires cleaning up.

projects/user_publication/gscholar.py
AnishReddyRavula: Good catch. Not sure how I missed this.

projects/user_publication/gscholar.py
Mark-Powers: Yes, though as I mentioned, we only have one publication row per publication? So there will usually be multiple sources per publication

AnishReddyRavula: Yes. I am making a change as mentioned by Adam to have PublicationSources as a model to accommodate the information about the name of the source the publication found in and the citation count.

projects/user_publication/gscholar.py
```diff
@@ -120,33 +124,83 @@
...
-            # this mostly never happens - if it does - log it
-            logger.info(f"cannot parse bibtex, so ignoring - {bibt}")
-            return
-        authors = authors_match.split(" and ")
+        authors = utils.decode_unicode_text(pub['bib']['author'])
+        # substitute all charachter from text except
+        # ',' , \w word charachter, \s whitespace charachter
+        # - any other charachter
+        # inspired from django.utils.text.slugify
+        authors = re.sub(r"[^\,\w\s-]", "", authors)
```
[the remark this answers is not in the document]
AnishReddyRavula: Few responses have unicode charachters encoded as ascii - for example Lo{\"}c

AnishReddyRavula:
- Created a new model PublicationSource to store information about citation count from each source and also if a publication is identified from a source(using our publication to project mapping and citation algorithm)
- Addressed other comments

projects/models.py
```diff
@@ -325,3 +318,34 @@ class Funding(models.Model):
...
+    GOOGLE_SCHOLAR = "google_scholar"
+
+    SOURCES = [
+        (USER_REPORTED, "User Reported"),
+        (SCOPUS, "Scopus"),
+        (SEMANTIC_SCHOLAR, "Semantic scholar"),
+        (GOOGLE_SCHOLAR, "Google Scholar"),
+    ]
+
+    publication = models.ForeignKey(Publication, related_name="source", on_delete=models.CASCADE)
```
super-cooper, line 337: `related_name` here should probably be "sources", since it's one-to-many

projects/user_publication/citation.py
```diff
@@ -40,33 +41,35 @@ def update_scopus_citation(pub, dry_run=True):
...
-        logger.info(
-            (
-                f"update scopus citation number for "
-                f"{pub.title} (id: {pub.id}) "
-                f"from {existing_scopus_cites} "
-                f"to {scopus_pub.citedby_count}"
-            )
-        )
+            existing_scopus_source.citation_count = scopus_pub.citedby_count
+            existing_scopus_source.save()
```
super-cooper, line 59: Database updates should be done inside `transaction.atomic()`. We don't do it in a lot of places in portal... but we should!

projects/user_publication/utils.py
```diff
@@ -27,10 +27,15 @@
...
-    LOG.info(f"Publication already exists - {pub.title} - updating other source")
-    setattr(pub, source, True)
-    pub.save()
+def add_source_to_pub(pub, source):
+    LOG.info(
+        f"Publication already exists - {pub.title}"
+        f" - adding other source - {source} - This check might be outdated"
+    )
+    source = pub.source.get_or_create(name=source)[0]
+    source.is_source = True
```
super-cooper, line 36: Is this ever false?

super-cooper (approved):
This is looking great now! Excellent job. 

I have a few questions and comments attached to the code. One other thing I'm wondering: Are we planning to add the code which automatically looks up citations when users submit publications at a later date?

projects/user_publication/utils.py
AnishReddyRavula: yes. When we only get citations count for a publication but are not able to find that publication referencing chameleon when we search in that source.

projects/models.py
```diff
@@ -325,3 +318,34 @@ class Funding(models.Model):
...
+    name = models.CharField(max_length=30, choices=SOURCES)
+    citation_count = models.IntegerField(default=0, null=False)
+    # if the publication identified by the source
+    # using our algorithm to find publications ref Chamaleon
+    is_source = models.BooleanField(default=False, null=False)
```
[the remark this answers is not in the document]
AnishReddyRavula: Aaah. Spent like 17 minutes for a better name XD

projects/admin.py
```diff
@@ -72,7 +67,44 @@ def get_form(self, request, obj=None, **kwargs):
...
+        publication = getattr(model, "publication", None)
+        return getattr(publication, "id", None)
+
+    def publication_title(self, model):
+        """Obtain the publication title attribute from the `publication` relation """
+        publication = getattr(model, "publication", None)
+        return getattr(publication, "title", None)
+
+
+class PublicationSourceAdmin(PublicationFields, admin.ModelAdmin):
```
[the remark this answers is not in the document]
AnishReddyRavula: I wanted to do it, but to test a few things my admin page is not working. So I made a standard thing for now

AnishReddyRavula:
> This is looking great now! Excellent job.
> 
> I have a few questions and comments attached to the code. One other thing I'm wondering: Are we planning to add the code which automatically looks up citations when users submit publications at a later date?

We will have to run `update_citation_numbers()` function periodically

projects/models.py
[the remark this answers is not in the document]
Mark-Powers: What about `found_by_algorithm` here? `is_source` is very unclear to me.

Mark-Powers (approved):
Looks good, I think `is_source` should be renamed though to clarify what it is, Adam and I were both confused by it.

PR #386, opened 2023-03-23
Add roles to artifact edit form

super-cooper:
Users will now be able to assign and unassign roles from users when editing their artifacts.

The roles function very similarly to the authors, where new input fields are added statically.

image

### DIFF
--- a/sharing_portal/templates/sharing_portal/edit.html
+++ b/sharing_portal/templates/sharing_portal/edit.html
@@ -40,9 +40,9 @@
         </fieldset>
         {% endfor %}
         <h4>Roles</h4>
-        Assign roles to various users, allowing them to collaborate on this artifact. <br><br>
-        The <b>collaborator</b> role allows users to edit artifact metadata, upload new versions, and share private artifacts. <br><br>
-        The <b>administrator</b> role allows users to control almost everything about the artifact.<br><br>
+        Assign roles to various users, allowing them to collaborate on this artifact. <br>
+        For more information on roles, see
+        <a href="https://chameleoncloud.readthedocs.io/en/latest/technical/sharing.html#trovi-roles">our documentation</a>.<br><br>
         {{ roles_formset.management_form }}
         {% for form in roles_formset.forms %}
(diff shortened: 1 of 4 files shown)


super-cooper:
[Portal - Trovi role permissions](https://trello.com/c/oJJyLaB9/2957-portal-trovi-role-permissions)

sharing_portal/templates/sharing_portal/edit.html
```diff
@@ -35,7 +35,17 @@ <h2 class="artifactTitle">Edit artifact</h2>
...
         {% for form in authors_formset.forms %}
-        <fieldset class="form-group form-inline">
+        <fieldset class="form-group form-inline" prefix="author">
+          {% bootstrap_form form show_label=False layout='inline' %}
+        </fieldset>
+        {% endfor %}
+        <h4>Roles</h4>
+        Assign roles to various users, allowing them to collaborate on this artifact. <br><br>
+        The <b>collaborator</b> role allows users to edit artifact metadata, upload new versions, and share private artifacts. <br><br>
+        The <b>administrator</b> role allows users to control almost everything about the artifact.<br><br>
```
Mark-Powers, line 45: maybe we should say what exactly this means? It's unclear to me what "almost everything" refers to. Maybe we should link to a readthedocs paragraph explaining what each role can do.

Mark-Powers (approved):
Looks pretty straightforward! This should probably be accompanied by a documentation paragraph too.

sharing_portal/templates/sharing_portal/edit.html
super-cooper: Agree it's not clear. I went with this phrasing because I couldn't think of a nice way to say "full control, except you can't get rid of the owner's admin role."

Mark-Powers: Maybe what is most clear would be to have a table for these capabilities in the docs we can link to, and update there if things change.

super-cooper: Good idea, done.

PR #388, opened 2023-03-29
Throttling pubs changes

AnishReddyRavula:
Changes include a new process for identifying the publications that were enabled by Chamaleon
- individual pub_import functions in scopus.py, semantic.py and google_scholar.py are only return the publication model with all the information from the source.
- - publications.py will get all the models from each source and decides what publications to save by appropriately flagging them of duplicates
- A sources model FK to the Publications model
- `flag_duplicates` to flag the publication model as duplicate and it upto the calling program to decide what to do with those duplicates.
- `review_duplicates` to review the duplicate flagged publications

Publication review process here - https://docs.google.com/document/d/18G9K9d769ztL1jNruJwiI9ncIzE4pY2ebVRvQXnc-gs/edit


projects/user_publication/publications.py
```diff
@@ -1,17 +1,73 @@
...
+                utils.save_publication(pub, source)
+
+        # Export the publication status report
+        utils.export_publication_status_run(
+            report_file_name, pub, author_usernames,
+            valid_projects, projects, reason_for_report
+        )
+
+
+def review_imported_publications():
```
super-cooper, line 71: Is this supposed to be implemented?

projects/admin.py
```diff
@@ -30,13 +29,11 @@ class PublicationAdmin(ProjectFields, admin.ModelAdmin):
         "doi",
         "link",
         "added_by_username",
-        "entry_created_date",
```
Mark-Powers, line 33: Why is this dropped? I think we should keep track of this still.

projects/admin.py
```diff
@@ -30,13 +29,11 @@ class PublicationAdmin(ProjectFields, admin.ModelAdmin):
...
         "status",
-        "approved_with",
```
Mark-Powers, line 35: I see no migration that converts this data to the charfield. Is there any data in this column or was it new (and so doesn't exist in prod)?

projects/models.py
```diff
@@ -333,12 +317,36 @@ class PublicationSource(models.Model):
...
     publication = models.ForeignKey(Publication, related_name="sources", on_delete=models.CASCADE)
     name = models.CharField(max_length=30, choices=SOURCES)
     citation_count = models.IntegerField(default=0, null=False)
+    entry_created_date = models.DateField(auto_now_add=True)
     # if the publication identified by the source
-    # using our algorithm to find publications ref Chamaleon
-    found_by_algorithm = models.BooleanField(default=False, null=False)
+    # using our algorithm to find publications ref Chamaeleon
+    is_found_by_algorithm = models.BooleanField(default=False, null=False)
+    is_cited = models.BooleanField(default=False, null=False)
```
Mark-Powers, line 347: Nit: the name is a bit confusing. I think `cites_chameleon` and similarly `acknowledges_chameleon` for the next field would be more clear.

projects/user_publication/deduplicate.py
```diff
@@ -1,53 +1,87 @@
...
-                if not dry_run:
-                    with transaction.atomic():
-                        pub1.status = Publication.STATUS_DUPLICATE
-                        pub1.save()
-                # Log the publications that have been flagged as duplicates
-                logger.info(f"{pub1.title} is flagged duplicate")
-                # break the inner loop and continue the outer loop to check duplicate for
-                # a new publication
-                break
+            if not PublicationUtils.is_pub_similar(pub1, pub2):
```
Mark-Powers, line 57: I think it would be clearer to have `if PublicationUtils.is_pub_similar(pub1, pub2):` and then moving the status update to inside the if, rather than use continue. It takes more thought to understand the negative condition + continue.

Mark-Powers (changes requested):
I had some important concerns about the data model, we want to make sure not to drop any information.

projects/admin.py
AnishReddyRavula: You are right, we should keep track of this. So I added this field in `PublicationSource` model so we can keep track of each source adding the publication

projects/admin.py
AnishReddyRavula: Yes. I should also have a migration/workflow to store the existing data in new model

AnishReddyRavula:
> Looks good! I'm still not sure after this review what exactly the "throttling phase" is. I've seen it mentioned a lot but never got an explanation. Were we just getting rate limited by the publication APIs?

Throttling comes after generating (getting all the pubs) which is to limit the publications that we consider as "used chameleon for research"

super-cooper:
> > Looks good! I'm still not sure after this review what exactly the "throttling phase" is. I've seen it mentioned a lot but never got an explanation. Were we just getting rate limited by the publication APIs?
> 
> Throttling comes after generating (getting all the pubs) which is to limit the publications that we consider as "used chameleon for research"

I'm still not sure I understand what throttling means in that context. Does this refer to filtering the projects? My understanding of "throttling" is a bottleneck on performance, rather than cleaning data.

projects/admin.py
AnishReddyRavula: Added migrations with data migration logic.

Mark-Powers (approved):
Looks good!

PR #389, opened 2023-03-30
Request to join project link

super-cooper:
This change introduces two new database models:

* `JoinLink` — Represents a secret link which PIs can send out. Clicking this link allows users to join a project.
* `JoinRequest` — Generated after a user clicks a `JoinLink` and confirms their intent to join a project. `JoinRequests` default to state `Pending`, and can be `Accepted` or `Rejected` by a membership manager. A user can only request to join a project one time.

`JoinLink`s are now embedded in the modal which pops up with the `Add multiple users` button. This will inform PIs of the links when they are looking for a method to add multiple users. 

When a user clicks a `JoinLink`, they are brought to a basic page which informs them that they are trying to join a particular project, and they can confirm their intent to join, or cancel. 

Screenshot from 2023-03-29 17-49-12

When the user clicks "Cancel," they are redirected to their project dashboard. When they click "Confirm", a `JoinRequest` is created, and an email is sent to the project's PI to inform them that someone has requested to join.

Project PIs are able to view join requests in their project dashboard in the same location as existing members and invites. Join requests come with a configuration button which allows PIs to accept or reject the requests. When they make a choice, the associated `JoinRequest` is updated, and an email is sent to the requesting user informing them of the PI's decision.

Screenshot from 2023-03-29 18-04-10


super-cooper:
> Does the gear icon appear for non pi users next the request in the membership list? It looks to me like it does, but since it's based on the other code, I'm unsure.

You're right, the gear would show up, but wouldn't be usable. I moved the guards out to fix it.

> How cumbersome do you think it is for the manager to go through each request? Since it's under the gear, it adds another click or two. I wonder if buttons that always show would be easier, but this might not matter in the first place.

I think it's cumbersome regardless, since the page has to refresh every time a request is decided. The only way to make this quicker would be to do it in javascript. If we wanted to put the accept/reject buttons outside of the gear wheel, I don't think it would fit very well into the existing table.

AnishReddyRavula:
Looks great!

AnishReddyRavula:
Looks great!

AnishReddyRavula (approved):
Looks great!

PR #390, opened 2023-04-05
Store duplicate pubs

AnishReddyRavula:
Ability to store the original reference based on which a publication was flagged as duplicate.
This can be used to get all related publications and maybe one of them has the correct data (with doi, bibtex etc)


projects/models.py
```diff
@@ -365,3 +365,15 @@ def __repr__(self) -> str:
             for ck in self.SOURCE_REPORT_FIELDS
         ]
         return "\n" + "\n".join(lines)
+
+
+class PublicationDuplicate(models.Model):
```
super-cooper, line 370: I think we could simplify this without a new model. We could just mark `Publication`s with status `'duplicate'` and then add a nullable field `duplicates`/`duplicate_of` which points to another publication.

super-cooper (commented):
The code looks good but I have one question about the new model.

projects/models.py
AnishReddyRavula: We can simply by adding a duplicate_of nullable field to the Publication model but the reason for using another model is to isolate the Publication model/table and not make many changes to it

super-cooper (approved):

Mark-Powers (approved):

PR #392, opened 2023-04-08
Used in research page

AnishReddyRavula:
Displays the user-uploaded and imported Publications in a tabular format. 
Also included is a change in data migration to change the status of imported publication status as 'SUBMITTED' instead of 'IMPORTED' as a source other than user-uploaded means that it is imported.

image


Mark-Powers (approved):
Looks good!

projects/templates/projects/chameleon_used_in_research.html
```diff
@@ -0,0 +1,38 @@
...
+
+  <h1>Chameleon Used In Research</h1>
+  <table id="example" class="table table-striped table-bordered align-middle" style="width:100%">
+    <thead>
+        <tr>
+            <th>Title</th>
+            <th>Year</th>
+            <th>Forum</th>
+            <th>Authors</th>
+            <th>Cites</th>
```
Mark-Powers, line 14: This should be "Cited by" I think, like how google scholar displays it. "Cites" is unclear whether that is how many papers this article cites, or how many papers cite this article.

PR #393, opened 2023-05-01
Bug fixes after pubs run

AnishReddyRavula:
- No changes to the chameleon used in the research view, but it is showing up in changes

- Fix in data migration to have the appropriate flag for `found_by_algo`
- Only approved_with 'Publication' will be `cited_by_chameleon`
- Renamed `reviewed` to `checked_for_duplicates`
- renamed `flag_duplicates` to `get_duplicate_pubs` so the caller can handle the duplicates
- No changes to the chameleon used in the research view, but it is showing up in changes
- In deduplicates - pick the right duplicate method ( flag 'preprint' as duplicate)
- Made the reviewing publications function more robust

Not sure why so many changes. But it is what it is


projects/user_publication/gscholar.py
```diff
@@ -204,8 +203,9 @@ def pub_import(
         cited_pubs = gscholar.get_cites(ch_pub, year_low=year_low, year_high=year_high)
         for cited_pub in cited_pubs:
             pub_model = gscholar.get_pub_model(cited_pub)
-            same_pub = utils.get_publication_with_same_attributes(pub_model, Publication)
-            if same_pub.exists():
-                utils.add_source_to_pub(same_pub.get(), PublicationSource.GOOGLE_SCHOLAR)
+            try:
+                pub_model.year = int(pub_model.year)
```
super-cooper, line 207: `Publication.year` is a non-null integer field. I would be very surprised if we found a publication without a year, but also, if there _isn't_ a valid year, saving this `Publication` will fail.

projects/user_publication/publications.py
```diff
@@ -28,6 +33,11 @@ def import_pubs(dry_run=True, file_name="", source="all"):
 
     # Process each publication
     for source, pub in pubs:
+        same_pubs = utils.get_publications_with_same_attributes(pub, Publication)
+        if same_pubs.exists():
```
super-cooper, line 37: This check is unnecessary. A loop through an empty `QuerySet` is a no-op.

projects/user_publication/publications.py
```diff
@@ -28,6 +33,11 @@ def import_pubs(dry_run=True, file_name="", source="all"):
 
     # Process each publication
     for source, pub in pubs:
+        same_pubs = utils.get_publications_with_same_attributes(pub, Publication)
+        if same_pubs.exists():
+            for same_pub in same_pubs:
+                utils.add_source_to_pub(same_pub, source)
+            continue
```
super-cooper, line 40: This `continue` is unnecessary as well. This is at the end of the loop anyway.

projects/user_publication/publications.py
```diff
@@ -93,48 +94,150 @@ def choose_approved_with_option():
...
+    email_choice = choose_email_approval_status()
+    if email_choice == 'approved':
+        with transaction.atomic():
+            user_reported_source.approved_with = PublicationSource.APPROVED_WITH_EMAIL
+            user_reported_source.save()
+            pub.status = Publication.STATUS_APPROVED
+            pub.save()
+    elif email_choice == 'rejected':
+        with transaction.atomic():
+            user_reported_source.approved_with = PublicationSource.APPROVED_WITH_EMAIL
```
super-cooper, line 140: This should not be set if the publication is rejected.

projects/user_publication/publications.py
```diff
@@ -93,48 +94,150 @@ def choose_approved_with_option():
...
+            user_reported_source.save()
+            pub.status = Publication.STATUS_REJECTED
+            pub.save()
+    else:
+        with transaction.atomic():
+            user_reported_source.approved_with = PublicationSource.APPROVED_WITH_PENDING_REVIEW
+            user_reported_source.save()
+
+
+def update_user_reported_status(pub, user_reported_source):
```
super-cooper, line 150: I'm not sure what this function is for. These publications are usually reviewed in the admin interface.

projects/user_publication/publications.py
```diff
@@ -93,48 +94,150 @@ def choose_approved_with_option():
...
+        for source in pub.sources.all():
+            print(f"with source: {source.__repr__()}\n")
         approval_needed_sources = pub.sources.filter(
             approved_with=PublicationSource.APPROVED_WITH_PENDING_REVIEW
-        )
-        if approval_needed_sources:
+        ).exclude(name=PublicationSource.USER_REPORTED)
+        user_reported_source = pub.sources.filter(
+            name=PublicationSource.USER_REPORTED,
+        ).first()
```
super-cooper, line 214: This call will fail if the filter is empty

projects/user_publication/deduplicate.py
```diff
@@ -11,9 +10,30 @@
...
+        # Check if dpub and pub2 are similar enough to be flagged as duplicates
+        if PublicationUtils.is_pub_similar(dpub, pub2):
+            original_pubs.append(pub2)
+    return original_pubs
+
+
 def get_duplicate_pubs(pubs=None):
     """
     returns duplicate publications in the database.
+    If passed pubs are not stored in database - this does not work.
```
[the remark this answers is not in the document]
AnishReddyRavula: My bad. I did not frame the doc right. If the passed pubs are not stored (saved) in the database the function fails with an error. Will correct it

projects/user_publication/gscholar.py
AnishReddyRavula: Yes, it will skip the publication. As for Gscholar, all we are doing is scraping. It sometimes may not have a year to the publication. In my runs, there is one publication for which the whole script will fail, and it's really annoying. So I included the catch here. Will log it and skip the publication.

projects/user_publication/publications.py
AnishReddyRavula: I would like to skip the publication if there is at least a single publication with the same attributes after updating all the same publications so need that check.

projects/user_publication/publications.py
AnishReddyRavula: yeah. It is confusing. The meaning of this is, it was tried for approval with email and then got rejected from email.
This helps us maintain a count of how many got rejected from email.

projects/user_publication/publications.py
AnishReddyRavula: I think the database I had a few unreviewed publications so I ended up implementing this. After this change, I don't think doing this admin interface is going to be easy. I will have to make changes admin interface to support this. Will create a Trello card for this.

AnishReddyRavula: I am removing this and also updated the admin interface to display source information allowing to make approvals on the publications page itself.

projects/user_publication/deduplicate.py
```diff
@@ -83,38 +83,37 @@ def flag_duplicate_in_whole_db():
...
+
+    duplicate_has_preprint = has_preprint(dpub)
+    original_has_preprint = has_preprint(opub)
+
+    if duplicate_has_preprint and original_has_preprint:
+        duplicate_pub = dpub
+        original_pub = opub
+    elif duplicate_has_preprint:
         duplicate_pub = dpub
         original_pub = opub
```
super-cooper: I'm confused by this. Both of these branches have the same body. Could we not just check 
```python
if duplicate_has_preprint:
    duplicate_pub = dpub
    original_pub = opub
else:
    duplicate_pub = opub
    original_pub = dpub
```

projects/user_publication/publications.py
super-cooper: I don't think we should be using the `approved_with` field to describe how a publication is rejected. I think this should be left null. 

If a publication is rejected, it's basically for two reasons:

1. It's very obvious (published in 1999 or something like that)
2. I emailed the PI and they said they didn't use Chameleon

projects/user_publication/publications.py
```diff
@@ -129,47 +129,26 @@ def choose_email_approval_status():
 
 def update_status_for_email_approval(pub, user_reported_source):
```
super-cooper, line 130: My comments below aside, what is this function for? As mentioned in the last review, publication reviews are usually done in the admin interface. Email reviews usually take time, and sometimes never get responses at all, so having a CLI for this seems cumbersome. Much easier to send the review and then click one button in the admin interface.

projects/user_publication/publications.py
```diff
@@ -104,7 +104,7 @@ def choose_approved_with_option():
     elif inp == "3":
         return PublicationSource.APPROVED_WITH_EMAIL
     elif inp == "4":
-        return 
+        return
     elif inp == "5":
         return PublicationSource.APPROVED_WITH_PENDING_REVIEW
```
super-cooper, line 109: Come to think of it, what is this status for? If the publication hasn't been reviewed, then it hasn't been approved. So how is something approved with a pending review?

projects/user_publication/publications.py
super-cooper: My bad 😅

projects/user_publication/publications.py
AnishReddyRavula: This is just to indicate that the publication source still needs approval.
I think it is better to remove this and add another field to indicate the source needs a review.

super-cooper (approved):
Nicely done on the new changes :D

PR #395, opened 2023-05-15
Chameleon used in research - Publication changes

AnishReddyRavula:
New workflow for storing publications from other sources
- Import publications from Gscholar
- Publication model changes to store source information in another model along with the flags (is_cited, is-acknowledged, etc)
- Classify if a publication used chameleon
 -  - If there is a citation to any of the Chameleon publications
 -   - If any of the authors are part of Chameleon
 -   - If the date of publication is after any of the allocations of the user
- Review imported publications - if they really use Chameleon
- Flag a publication with its source if the publication already exists
- Flag a duplicate publication and send an email to the operator about the duplicate when the user uploads
- Store the duplicate publications in a table so we can track duplicates.


super-cooper (approved):
Let it rip! :shipit: 

We can start the deployment process tomorrow morning.

PR #398, opened 2023-05-23
Fix issues with Daypass error handling

super-cooper:
Daypass has been mostly broken since we overhauled Trovi. This patch fixes various errors which prevented Daypass from working.

### DIFF
--- a/projects/tasks.py
+++ b/projects/tasks.py
@@ -102,9 +102,10 @@ def handle_too_many_daypass_users(artifact_uuid):
     admin_token = trovi.get_client_admin_token()
     artifact = trovi.get_artifact(admin_token, artifact_uuid)
     artifact_title = artifact["title"]
-    admin_usernames = trovi.get_administrators(artifact)
-    keycloak_client = KeycloakClient()
-    admin_users = [keycloak_client.get_user_by_username(u) for u in admin_usernames]
+    project = trovi.get_linked_project(artifact)
+    if not project:
+        return
+    managers = [u.email for u in get_project_membership_managers(project)]
 
     subject = "Pause on daypass requests"
     help_url = "https://chameleoncloud.org/user/help/"
@@ -127,7 +128,7 @@ def handle_too_many_daypass_users(artifact_uuid):
     send_mail(
         subject=subject,
         from_email=settings.DEFAULT_FROM_EMAIL,
-        recipient_list=[u["email"] for u in admin_users],
+        recipient_list=managers,
         message=strip_tags(body),
         html_message=body,
     )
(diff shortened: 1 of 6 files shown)


msherman64 (approved):

Mark-Powers (approved):
Maybe I am missing context, but the reasoning for using managers of a project rather than anyone is that preserves a chain of responsibility. It was

1. user creates an artifact
2. user assigns it to project X
3. the reproducibility project Y is created with same PI as X
4. the PI can add managers to Y to deal with the requests

but the changes here change that, and it seems like non-PIs/manageres are giving access to chameleon. 

Additionally, the idea behind `list_daypass_requests` being artifact agnostic was a user can review all requests across different artifacts easier. Maybe that is too confusing.

Other than this policy issue I mention, the code looks fine.

super-cooper:
Ah I see what you're saying. This is a problem. The whole reason for this change was not a policy one, but because daypass was not functioning. I think the policy you're describing here makes sense, but Daypass was having issues with artifacts that didn't have a `linked_projects`. I think, in general. there are problems with centering the policy around Chameleon as Trovi expands into other testbeds as well. 

The Daypass reviews are currently limited to Chameleon users. Do you think we should instead limit it only to Chameleon users who are also PIs of at least one project? And then we could make all those users managers of the repro project? That's not exactly ideal, but it does at least carry over some of the old policy.

Mark-Powers:
@super-cooper 
>  I think, in general there are problems with centering the policy around Chameleon as Trovi expands into other testbeds as well.

That is the goal of using `linked_projects`, as it keeps it fairly untied to Chameleon. If there is no linked project, there should not be a "request daypass" button.

> The Daypass reviews are currently limited to Chameleon users. Do you think we should instead limit it only to Chameleon users who are also PIs of at least one project? And then we could make all those users managers of the repro project? That's not exactly ideal, but it does at least carry over some of the old policy.

We were intending to use membership of the linked project so that managers and the PI of it can review daypasses. I don't think we should require PI status, as most PIs are pretty busy. I'm not sure why this isn't working with the additional of roles, as there shouldn't be anywhere it checks against trovi permissions.

Mark-Powers (approved):
Looks good! Since this came up, it seems like somewhere `linked_projects` is not set right, and I'm not sure why it would be the case it suddenly stopped working.

super-cooper:
> it seems like somewhere `linked_projects` is not set right, and I'm not sure why it would be the case it suddenly stopped working.

Yeah, it's definitely strange. Could have been behavior from an older version of Trovi, since the current form requires you to link a project. This would have been good to have anyway because other testbeds may have different sharing models than we do.

PR #399, opened 2023-06-20
Add featured artifacts to top of Trovi homepage

super-cooper:
This feature is in its most basic configuration. The featured artifacts are stored in a special table in the datbase, and boosted to the top of the Trovi homepage. 

New featured artifacts are added by using the Featured Artifacts admin page and pasting a UUID in the "New Featured Artifact" form. If the UUID does not map to a real Trovi artifact, the UUID will be ignored when the Trovi homepage is rendered.

I considered making a nicer admin page with a dropdown populated by `ListArtifact`, but the amount of work it required seemed not worth it for a page that will be used rarely and only by operators.

image

image


Mark-Powers (approved):
👍

super-cooper:
[Show the best artifacts at the top of the trovi page](https://trello.com/c/fwJPkTzZ/3013-show-the-best-artifacts-at-the-top-of-the-trovi-page)

AnishReddyRavula (approved):
Looks great

PR #402, opened 2023-07-11
fix enforcement exception message

AnishReddyRavula:
`LeasePastExpirationError` is not sending the right message on raising the exception
Fix `EnforcementException` to use the message of subclass

when a `message` is not passed, it will be `None` and it overwriting `self.message`. This fix will only overwrite it if `message` is passed with contents.

### DIFF
--- a/balance_service/enforcement/exceptions.py
+++ b/balance_service/enforcement/exceptions.py
@@ -5,7 +5,6 @@ LOG = logging.getLogger(__name__)
 
 class EnforcementException(Exception):
     """Base enforcement Exception.
-
     To correctly use this class, inherit from it and define
     a 'msg_fmt' and 'code' properties.
     """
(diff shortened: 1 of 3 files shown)


AnishReddyRavula:
> What exactly was the bug with the exception code? It seems to be doing the same exact stuff with your changes as before.

Updated the PR desc. 

>when a message is not passed, it will be None and it overwriting self.message. This fix will only overwrite it if message is passed with contents.

Mark-Powers:
Yes, but if message is None, it should then be overwritten with `self.message = self.msg_fmt % kwargs`, right?

AnishReddyRavula:
> Yes, but if message is None, it should then be overwritten with `self.message = self.msg_fmt % kwargs`, right?

But `msg_fmt` is "Not authorized" and `kwargs` is `{}` so it is escaping the actual message

Mark-Powers:
I see, so it's overwriting the message of `LeasePastExpirationError`? I think we should update that class so `msg_fmt` is "Cannot create reservation beyond allocation expiration date" instead, so that is always used.

Mark-Powers (approved):

PR #407, opened 2023-07-31
Changes to show Reproducible badge in trovi page

AnishReddyRavula:
- Create a Model `Artifactbadge` to hold all the details - name of the badge, who requested, when decision was made etc
- Show the logo in artifact list page
- Filter by badge similar to filter with tag
- Show the badge and chameleon logo in detail page

### DIFF
--- a/sharing_portal/admin.py
+++ b/sharing_portal/admin.py
@@ -1,7 +1,9 @@
 from django.contrib import admin
 
-from .models import DaypassRequest, FeaturedArtifact
+from .models import DaypassRequest, FeaturedArtifact, ArtifactBadge
 
 admin.site.register(DaypassRequest)
 
 admin.site.register(FeaturedArtifact)
+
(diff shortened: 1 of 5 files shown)


AnishReddyRavula:
Need to test in dev. Would like your input on the design.

super-cooper (approved):
I actually like this idea more than the one I had for privileged tags in Trovi. I think we should consider making one of these for the Chameleon logo and retiring the Chameleon tag so people stop misusing it.

Mark-Powers (approved):
This is great! I think perhaps the logo should say "Reproducible" next to it to make things more clear.

AnishReddyRavula:
> This is great! I think perhaps the logo should say "Reproducible" next to it to make things more clear.

Kate suggested that we show a description of the badge and also a redirect link to a form-  application for the badge

PR #410, opened 2023-08-08
Feedback button which redirects to github issues

AnishReddyRavula:
This is when artifact contains github url

image

### DIFF
--- a/sharing_portal/views.py
+++ b/sharing_portal/views.py
@@ -616,6 +616,9 @@ def artifact(request, artifact, version_slug=None):
             messages.error(request, message)
 
     git_content = [method for method in access_methods if method["protocol"] == "git"]
+    if len(git_content) > 0:
+        # git repository is /<repository>.git format
+        feedback_url = git_content[0]['remote'].replace(".git", "/issues")
     http_content = [method for method in access_methods if method["protocol"] == "http"]
 
     # has_reproducible_badge, is_chameleon_supported needs to be populated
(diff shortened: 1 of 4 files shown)


sharing_portal/views.py
```diff
@@ -632,6 +635,7 @@ def artifact(request, artifact, version_slug=None):
         "show_launch": show_launch,
         "git_content": git_content,
         "http_content": http_content,
+        "feedback_url": feedback_url,
```
super-cooper, line 638: This will be an unresolved symbol if the artifact doesn't use git.

super-cooper (changes requested):
Seems fine to me as long as we're cool with them getting a 404 on certain repos. I believe we already had that discussions.

Requesting changes because there is a critical bug, but other than that I approve.

super-cooper:
Also, in theory, an artifact can use different git repos for different versions. Are we cool always taking the first one?

super-cooper (approved):

sharing_portal/views.py
```diff
@@ -616,6 +616,10 @@ def artifact(request, artifact, version_slug=None):
             messages.error(request, message)
 
     git_content = [method for method in access_methods if method["protocol"] == "git"]
+    feedback_url = ""
+    if len(git_content) > 0:
+        # git repository is /<repository>.git format
+        feedback_url = git_content[0]["remote"].replace(".git", "/issues")
```
Mark-Powers, line 622: This should explicitly check for github, and may need to extract the username/repo name to be correct, as the url might be `git://` too. For trovi, we use a library to do so: https://github.com/ChameleonCloud/trovi/blob/main/trovi/storage/backends/git.py#L57

Mark-Powers (changes requested):
We need to be a bit cleverer about parsing the version url.

Mark-Powers (approved):
This looks great! I like the tests too! Did the tests require any special configuration to run? If so, it would be useful to add a short section to the readme with instructions.

PR #411, opened 2023-08-09
Add allocation expiration/activation logging

super-cooper:
The allocation system now logs when allocations are activated and deactivated.

### DIFF
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -144,10 +144,17 @@ def _deactivate_allocation(alloc):
     else:
         balance = balance[0]
         alloc.su_used = balance["used"]
+    LOG.info(f"Deactivating allocation {alloc.id} for project {alloc.project.charge_code}")
     alloc.status = "inactive"
     alloc.save()
-    KeycloakClient().update_project(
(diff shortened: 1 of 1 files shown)


allocations/tasks.py
```diff
@@ -247,12 +256,20 @@ def active_approved_allocations():
...
+                LOG.info(f"Activating allocation {alloc.id} for project {charge_code}")
                 alloc.status = "active"
                 alloc.save()
-                KeycloakClient().update_project(
+                keycloak_client = KeycloakClient()
+                keycloak_client.update_project(
                     charge_code, has_active_allocation="true"
                 )
 
+                updated_project = keycloak_client._lookup_group(charge_code)
```
Mark-Powers, line 267: Thoughts on just moving the alloc.save() line into an an `if` using this data?

Mark-Powers (approved):

allocations/tasks.py
super-cooper: I'm nervous about messing with behavior until we determine exactly what's going on. This whole thing is so fickle I don't want to potentially make things worse.

super-cooper:

PR #412, opened 2023-08-09
Add consistency check to allocation tasks

super-cooper:
After Allocation statuses are updated, there is an additional check for allocations that have inconsistent status between Portal and Keycloak

### DIFF
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -9,6 +9,7 @@ from celery.decorators import task
 from django.conf import settings
 from django.core.mail import send_mail
 from django.db import transaction
+from django.db.models import Q
 from django.forms.models import model_to_dict
 from django.utils import timezone
 from django.utils.html import strip_tags
@@ -273,6 +274,8 @@ def activate_expire_allocations():
     deactivate_multiple_active_allocations_of_projects()
     # activate allocations
     active_approved_allocations()
+    # check consistency between allocation system and Keycloak
+    check_keycloak_consistency()
 
 
 def _fill_charge_tmp_resource_ids():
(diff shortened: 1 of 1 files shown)


allocations/tasks.py
```diff
@@ -346,14 +390,16 @@ def check_charge():
             openstack_endtime = openstack_r.get("end_time")
             portal_endtime = portal_r.get("end_time").strftime(utils.DATETIME_FORMAT)
             if openstack_endtime != portal_endtime:
-                LOG.error(
-                    f"{resource_id} at {region} has incorrect end time! "
-                    f"openstack={openstack_endtime}, portal={portal_endtime}"
-                )
+                pass
```
Mark-Powers, line 393: I presume these were just spamming the logs?

Mark-Powers (approved):

super-cooper:

allocations/tasks.py
super-cooper: Yes. The logs are not helpful, but the checks have to be left in because when we removed them everything broke.

allocations/tasks.py
```diff
@@ -305,6 +308,47 @@ def _fill_charge_tmp_resource_ids():
...
+
+    keycloak_client = KeycloakClient()
+    groups = keycloak_client._project_admin()
+    groups_url = groups._client.get_full_url(
+        groups.get_path("collection", realm=keycloak_client.realm_name)
+    )
+
+    active_groups = {
+        group.get("name")
+        for group in groups._client.get(url=groups_url, has_active_allocation="true")
```
msherman64, line 332: I don't fully understand which API is being called here. To note, as per https://www.keycloak.org/docs-api/18.0/rest-api/#_groups_resource

the call to `GET /{realm}/groups` has the caveat "Only name and ids are returned.", and doesn't include attributes.
We might be able to avoid this by specifying `briefRepresentation=false` as per the above docs.

msherman64 (approved):
LGTM, depending on whether that API call actually returns attributes for each group.

allocations/tasks.py
super-cooper: Fixed. Turns out, the kwargs passed to that function do nothing and just silently fail :)))

PR #413, opened 2023-08-11
Correct Keycloak consistency errors in Allocation tasks

super-cooper:
In the code where consistency errors were reported, they are now also corrected.

### DIFF
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -13,6 +13,8 @@ from django.db.models import Q
 from django.forms.models import model_to_dict
 from django.utils import timezone
 from django.utils.html import strip_tags
+from keycloak.exceptions import KeycloakClientError
+
 from allocations.models import Allocation, Charge
 from balance_service.utils.su_calculators import project_balances
 from balance_service.enforcement.usage_enforcement import TMP_RESOURCE_ID_PREFIX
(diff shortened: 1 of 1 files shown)


Mark-Powers (approved):
Is there any way this could collide with the normal allocations task?

super-cooper:
> Is there any way this could collide with the normal allocations task?

If I'm understanding your concern correctly, I don't believe so. This code is only called after all the other allocation logic is executed, so they wouldn't be executed at the same time.

```python
@task
def activate_expire_allocations():
    # expire allocations
    expire_allocations()
    # check projects with multiple active allocations
    deactivate_multiple_active_allocations_of_projects()
    # activate allocations
    active_approved_allocations()
    # check consistency between allocation system and Keycloak
    check_keycloak_consistency()
```

super-cooper:

allocations/tasks.py
```diff
@@ -348,28 +350,47 @@ def check_keycloak_consistency():
...
+            for inactive in ("false", None)
+        )
     }
 
     LOG.info(f"CONSISTENCY: {len(active_groups)} active groups in Keycloak")
     LOG.info(f"CONSISTENCY: {len(inactive_groups)} inactive groups in Keycloak")
 
     for project in active_projects:
-        if project not in active_groups:
+        if project in inactive_groups:
```
msherman64, line 370: @super-cooper 
Just to comment, this would be cleaner if e.g. both `active_projects` and `inactive_groups` were `set()`s, rather than 1 set and one dict, so we could use set notation for the intersections.

active_in_portal_inactive_in_keycloak = active_projects-active_groups
inactive_in_portal_active_in_keycloak = inactive_projects-inactive_groups

super-cooper: They all appear to be set comprehensions, so they could be combined as such.

PR #417, opened 2023-08-21
override delete publication method to change the status of the pub to be deleted

AnishReddyRavula:
override delete publication method to change the status of the pub to be deleted

show publication status on the user list publications page

add blank=True to user publication source approved with - so the admin form does not have it as required

trello task - https://trello.com/c/HtV9BDCH/622-publications-admin-page-is-missing-columns


super-cooper:
I'm not sure I agree with overriding the delete function on the model. My assumption when the publication changes were originally made was that I would be able to set a publication to "REJECTED" in the admin interface. This override prevents us from actually deleting any publications.

super-cooper (approved):

PR #422, opened 2023-09-18
PI delegate users - Managers can request for allocation too

AnishReddyRavula:


Mark-Powers (approved):
This looks good! I was surprised at how short this is. The create allocation button shows up for everyone in the project, but only PI can actually use it?

AnishReddyRavula:
> This looks good! I was surprised at how short this is. The create allocation button shows up for everyone in the project, but only PI can actually use it?

The "create" can only be seen by the PI.
But "renew/entend" button is seen by only PIs and managers.

PR #423, opened 2023-09-29
Invitation admin features

AnishReddyRavula:
display invite url in admin page of invitation

Resend invitation to multiple invitations (action)

This change so when a user did not receive a invitation for somereason, we can initate a resend so they can use it

use get_invite_url from invitation models in sharing views

image
image


Mark-Powers (approved):
This looks great!

PR #425, opened 2023-10-05
Donot show private icon if doi

AnishReddyRavula:
Do not show a private icon if any of the versions have a DOI

add has_doi field to the artifact by looking at versions for doi


Mark-Powers (approved):
Looks great!

PR #426, opened 2023-10-06
Ticket detail page, text is not wrapped - this fixes it

AnishReddyRavula:
pre tag does not support text wrapping natively, adding a style to it to allow text wrapping

Before
image
 After
image


Mark-Powers (approved):
Looks great!

PR #427, opened 2023-10-06
Use reCaptchaV3 so users don't have to use checkbox

AnishReddyRavula:
We need to have the keys configured for v3 in the env, for this to work


Mark-Powers (approved):
:shipit:

PR #429, opened 2023-11-02
Include publication admin url in the new pub uploaded notification email

AnishReddyRavula:


Mark-Powers (approved):
I assume this is just using the `__repr__` representation of publication? That seems to include the url, but did you check this isn't using `__str__`? I'm not certain of the default python behavior for string format.

AnishReddyRavula:
> I assume this is just using the `__repr__` representation of publication? That seems to include the url, but did you check this isn't using `__str__`? I'm not certain of the default python behavior for string format.

You are right. I should be using `__repr__()` explicitly. As I was using for the duplicate notification

PR #431, opened 2023-11-13
Add the badge filter to search the DOM instead of making another request

AnishReddyRavula:
- Making another request is very slow as it involves reloading the whole webpage to get the query filter from the URL
 - Using a button tag with a badge image inside it

Not a huge visual change, it is there if you look for it. 

Before 
image

After 
image


Mark-Powers (approved):
Nice simple change!

PR #432, opened 2023-11-14
Allocation tests

JOUNAIDSoufiane:
Added testing for allocations tasks and added instructions on how to run tests that require a DB connection

### DIFF
--- a/chameleon/settings.py
+++ b/chameleon/settings.py
@@ -145,7 +145,7 @@ INSTALLED_APPS = (
     "djangocms_picture",
     "djangocms_snippet",
     "djangocms_video",
-    "reversion",
+    # "reversion",
     ##
     # contrib
     #
@@ -257,6 +257,9 @@ if os.environ.get("DB_NAME"):
             "PORT": os.environ.get("DB_PORT"),
             "USER": os.environ.get("DB_USER"),
             "PASSWORD": os.environ.get("DB_PASSWORD"),
+            "TEST": {
+                'NAME': os.environ.get("DB_NAME") + "_test",
+            },
         },
     }
 else:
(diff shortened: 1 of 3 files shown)


chameleon/settings.py
```diff
@@ -257,6 +257,9 @@
             "PORT": os.environ.get("DB_PORT"),
             "USER": os.environ.get("DB_USER"),
             "PASSWORD": os.environ.get("DB_PASSWORD"),
+            "TEST": {
```
Mark-Powers, line 260: Is this built into the django database connection spec or something? I can't find reference to it.

Mark-Powers (approved):
This looks good! Just a few minor comments + can you fix the formatting issues the bot commented?

Regarding deploying, since this doesn't actually change the running code, it should be fine to deploy (or just wait until we deploy something else). We should set up this as a github action eventually, but for now we can at least test it while developing locally.

AnishReddyRavula (approved):
Like the tests! Especially the setUp and tearDown. 

I have a question, does all the other tests in portal code work with this change? if not please include in the readme which tests work or which doesn't.

chameleon/settings.py
JOUNAIDSoufiane: Yes, this is mentioned in [django docs](https://docs.djangoproject.com/en/4.2/topics/testing/overview/)

> The default test database names are created by prepending test_ to the value of each [NAME](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-NAME) in [DATABASES](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATABASES). When using SQLite, the tests will use an in-memory database by default (i.e., the database will be created in memory, bypassing the filesystem entirely!). The [TEST](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATABASE-TEST) dictionary in [DATABASES](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATABASES) offers a number of settings to configure your test database. For example, if you want to use a different database name, specify [NAME](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-TEST_NAME) in the [TEST](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATABASE-TEST) dictionary for any given database in [DATABASES](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-DATABASES).

PR #433, opened 2024-01-24
Allocations admin view (read only)

AnishReddyRavula:
This does not replace the existing allocatins approval page

Read only and cannot be used to change details

This is done so project manager can look at allocations and use that data to get insights into them

### DIFF
--- a/allocations/admin.py
+++ b/allocations/admin.py
@@ -6,7 +6,7 @@ class AllocationAdmin(admin.ModelAdmin):
     def has_change_permission(self, request, obj=None):
         return False
 
-    list_display= (
+    list_display = (
         'project',
         'status',
         'date_requested',
--- a/allocations/admin.py
+++ b/allocations/admin.py
@@ -1,17 +1,57 @@
+from django.conf import settings
 from django.contrib import admin
-from .models import Charge, Allocation
+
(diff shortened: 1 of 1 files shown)


Mark-Powers (approved):
This looks good!

marcwitasee:
@AnishReddyRavula I need information about the project content as well. For example, the description and the title of the project. I also would need to see the allocation justification and whether the allocation was approved as a renewal, new allocation, or a recharge. Is it possible to include that information in this view as well? Oh and also some information about the requester, name, email, affiliation.

AnishReddyRavula:
> @AnishReddyRavula I need information about the project content as well. For example, the description and the title of the project. I also would need to see the allocation justification and whether the allocation was approved as a renewal, new allocation, or a recharge. Is it possible to include that information in this view as well? Oh and also some information about the requester, name, email, affiliation.

- Renewal or new or recharge information cannot be shown as they are just placeholders on the other page to show the appropriate text, may be you can understand if that is new/renewal/recharge by looking at the Decision summary.
- I can do the description and title of the project. 
- I added the PI's name email and institution. But let me know if you also need requesters' to be seen on the same page. Requester will be mostly PI but can also be PI delegate. requesters name, username will be shown though

PR #434, opened 2024-02-02
Added restart=always directive to celery docker-compose entry

JOUNAIDSoufiane:
This change comes after a portal outage where celery did not come back on and new approved user projects were not moved to the active state.


Mark-Powers (approved):
This is fine to change but this is the local dev way we deploy it. The production docker-compose file is located in [portal-camino](https://github.com/ChameleonCloud/portal-camino/).

PR #437, opened 2024-03-08
Allow setting default and individual SU budgets for project members

AnishReddyRavula:
This commit introduces the functionality to set default and individual SU budgets for project members. It includes the following changes:

- Added models ChargeBudget to track individual SU budgets for users in projects.
- Added `default_su_budget` to Project model to enforce the budget for all other members
- Implemented set_budget_for_user_in_project function to set SU budgets for individual users in projects.
- Added form inputs to allow setting default SU budgets for all non-manager users in a project.
- Modified the view_project template to display current SU usage and budget inputs for project members.
- Implemented backend logic to handle setting default and individual SU budgets in the view_project view.

These changes enable project managers to manage SU budgets more effectively within projects.

[Docs PR](https://github.com/ChameleonCloud/docs/compare/project-su-budget?expand=1)


allocations/models.py
```diff
@@ -100,3 +104,49 @@ class Charge(models.Model):
...
+            return 0
+
+        charges = Charge.objects.filter(allocation=allocation, user=self.user)
+
+        # Avoid doing the calculation if there are no charges
+        if not charges.exists():
+            return 0
+        microseconds_per_hour = 1_000_000 * 3600
+        return charges.annotate(
+            charge_duration=ExpressionWrapper(
```
Mark-Powers, line 142: Could this have some comments to explain? It surprises me that this is anything but a sum of all of the charges. I'm especially confused why microseconds per hour comes up.

balance_service/enforcement/usage_enforcement.py
```diff
@@ -132,6 +132,19 @@ def get_balance_service_version(self, data):
         else:
             return 2
 
+    def _check_usage_against_user_budget(self, user, allocation, new_charge):
+        user_budget = ChargeBudget.objects.get(user=user, project=allocation.project)
+        left = user_budget.su_left()
+        if left - new_charge < 0:
```
Mark-Powers, line 138: I think `if left < new_charge` would be more clear.

projects/templates/projects/view_project.html
```diff
@@ -146,92 +146,174 @@ <h3>Project Members</h3>
...
-            <button type="submit" class="btn btn-xs btn-success btn-block" name="accept_join_request">Accept</button>
-            <button type="submit" class="btn btn-xs btn-danger btn-block" name="reject_join_request">Reject</button>
+            <div class="col-xs-4 col-sm-4 col-lg-4">
+              <div class="row">
+                <div class="col-xs-5 col-sm-5 col-lg-5">
+                  <b>Usage</b>: <p style="display: inline-block">{{u.su_used}}/</p>
+                  <p class="su-budget-display" style="display: inline-block">{{ u.su_budget }}</p>
+                </div>
+                <div class="col-xs-5 col-sm-5 col-lg-5">
+                  <input
```
Mark-Powers, line 243: How usable is this? It seems like it would be tricky to precisely set the slider. Also steps of 1000 is a lot, since our allocations only have 20000 SUs, and a class might have more than 20 people.

Mark-Powers: Maybe 500 would be sufficient step.

projects/views.py
```diff
@@ -401,6 +415,32 @@ def view_project(request, project_id):
...
+                budget_user, budget_user_project, request.POST["su_budget_user"], enforced_by
+            )
+            messages.success(
+                request,
+                (
+                    f"SU budget for user {budget_user.username} "
+                    f"is currently set to {charge_budget.su_budget}"
+                )
+            )
+        elif "default_su_budget" in request.POST:
```
Mark-Powers, line 432: I'm not sure this works, because if anyone joins the project late, they won't get a budget.

projects/views.py
```diff
@@ -312,6 +312,15 @@ def notify_join_request_user(django_request, join_request):
         )
 
 
+def set_budget_for_user_in_project(user, project, target_budget, enforced_by):
+    charge_budget, created = ChargeBudget.objects.get_or_create(
```
Mark-Powers, line 316: I'm forgetting how get_or_create works, but will this create a new object if a different manager sets the budget, since enforced_by will be different?

Mark-Powers (changes requested):
I have a few concerns about this implementation:
1. I think enforced_by should be removed, as I don't think it's ever exposed to users or would be particularly useful.
2. Projects should probably have a "default SU budget" field, and inside `_check_usage_against_user_budget` , if there is no ChargeBudget, it should instead use the project's default SU budget.

AnishReddyRavula:
> I have a few concerns about this implementation:
> 
> 1. I think enforced_by should be removed, as I don't think it's ever exposed to users or would be particularly useful.
> 2. Projects should probably have a "default SU budget" field, and inside `_check_usage_against_user_budget` , if there is no ChargeBudget, it should instead use the project's default SU budget.

Nice. These are very useful comments. Thanks. I will work on it

projects/views.py
```diff
@@ -401,6 +419,30 @@ def view_project(request, project_id):
...
+                request,
+                (
+                    f"SU budget for user {budget_user.username} "
+                    f"is currently set to {request.POST['su_budget_user']}"
+                )
+            )
+        elif "default_su_budget" in request.POST:
+            portal_project.default_su_budget = request.POST["default_su_budget"]
+            portal_project.save()
+            for u in users:
```
Mark-Powers, line 437: I think for consistency, this should be removed to the balance service check. That way, all of the chargebudget objects are created in the same place, when the user first tries to use SUs.

projects/views.py
```diff
@@ -530,6 +572,18 @@ def view_project(request, project_id):
...
+            try:
+                su_budget = ChargeBudget.objects.get(
+                    user=portal_user, project=budget_project
+                )
+            except ChargeBudget.DoesNotExist:
+                su_budget_value = current_allocation_su_allocated
+            else:
+                su_budget_value = su_budget.su_budget
+            user["su_budget"] = int(su_budget_value)
+            user["su_used"] = ChargeBudget(
```
Mark-Powers, line 584: It's a little confusing that this creates a ChargeBudget object, since it's not actually getting it from the database? Insteade maybe `current_usage` should be a static/util method somewhere?

Mark-Powers (changes requested):
A few minor comments to make things more consistent.

This needs a docs PR too.

Mark-Powers (approved):
Looks good!

PR #438, opened 2024-03-19
Pass when user is a manager and does not have a budget created

AnishReddyRavula:
This occurs, when user role needs to be changed to 'manager' and they currently do not have ChargeBudget set.

If the charge budget for a user that needs to be a manger is not set, ignore the exception and continue with the request/response


Mark-Powers (approved):
Code change looks OK, but probably should keep the `restart: always` in the compose file.

PR #439, opened 2024-03-22
Fix daypass projects to use daypass tag

AnishReddyRavula:
When creating daypass project users encountered error like this

`TypeError: int() argument must be a string, a bytes-like object or a number, not 'Tag'`

This issue started after this update - https://github.com/ChameleonCloud/portal/commit/f1ec9cc86c707cdaa8a382190b47c96d5286c32a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         if: github.event_name != 'pull_request'
         uses: sigstore/cosign-installer@main
         with:
-          cosign-release: 'v2.0.2' # optional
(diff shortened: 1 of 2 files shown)


Mark-Powers (approved):
Thanks! Were you able to test this works? I know creating new projects on dev is tricky. This looks like a correct fix based on the error message.

AnishReddyRavula:
> Thanks! Were you able to test this works? I know creating new projects on dev is tricky. This looks like a correct fix based on the error message.

I have not tested it yet. But as you said, this seems like a valid fix and straightforward. It is ridiculous how I did not see this earlier.

PR #444, opened 2024-05-01
Add max paramter to membership list

Mark-Powers:
By default, keycloak has:

Query param max: Maximum results size (defaults to 100)


msherman64 (approved):
Good workaround for now.

Adding the API Ref to the docs here: https://www.keycloak.org/docs-api/24.0.2/rest-api/index.html#_users

and a TODO, when we next revisit this we should make sure we use a method that handles pagination.

PR #448, opened 2024-06-26
Add check to refetch project membership if changed

Mark-Powers:
If a user was added/removed or a request accepted, the membership list would not show updates. This is especially evident with the new membership count being displayed.


JOUNAIDSoufiane (approved):
LGTM

PR #451, opened 2024-09-23
Bugfix/unittests

Mark-Powers:


allocations/migrations/0007_charge.py
```diff
@@ -70,6 +70,10 @@ def _str_to_localtime(utctimestr):
 
 
 def init_charge_from_openstack_db(apps, schema_editor):
+    # No initial charges in test DB
```
msherman64, line 73: Is this just to skip migrations when running in CI?
Do we know why this needs to access an external resource to run a migration anyway?

Mark-Powers: `settings.TESTING` is added in this PR, and is true when the code is run with `./manage.py test`.

This code initially loaded existing reservations at openstack sites into the charges database table.

balance_service/tests.py
```diff
@@ -0,0 +1,661 @@
...
+                                    "extra": {"availability_zone": "az1"},
+                                    "su_factor": update_su_factor,
+                                }
+                                for i in range(update_allocations_per_res)
+                            ],
+                        }
+                        for j in range(update_reservations)
+                    ],
+                },
+            }
```
msherman64, line 265: to comment, at a glance it's not clear to me which values are actually important, and which are just placeholders. Is this basically generating test data to help parametrize other tests?

msherman64 (approved):
Absolutely better than before, and I'm really happy to see where we explicitly need to mock calls to keystone and keycloak in order for it to work.

comments so far are mostly pointing at potential tech debt, but this seems good to me.
Are there any particularly annoying test cases where we should add comments describing the intended behavior?

balance_service/tests.py
Mark-Powers: This is generating blazar API data to test the methods that our API view calls. https://docs.openstack.org/blazar/2024.1/admin/usage-enforcement.html

The most relevant parts are # of reservations, # of allocations in each reservation, and su factor, along with the lease duration.

Mark-Powers:
Also calling out this functional change: 
https://github.com/ChameleonCloud/portal/blob/d98dcec52600a4eab781adfb901ee63c2c43b8ba/balance_service/enforcement/usage_enforcement.py#L206-L215

Before, it always used a temp ID, as the blazar enforcement API does not give reservation IDs on create. Now, this first looks for a real ID, and falls back on a temp ID. A celery task renames the temp IDs. I'm working to add a real ID to this API call in blazar. The temp IDs are not reliable, and cause errors when a user extends their lease.

PR #452, opened 2024-09-24
Delete stylelint reporter

Mark-Powers:
This runner doesn't seem to work, and not needed due to the low volume of CSS that we write for portal.


msherman64 (approved):
fine by me, I'd only ever added all the linters to see what's even useful.

PR #454, opened 2024-09-30
Refactor user permissions

Mark-Powers:
This fixes a bug where managers could do what they were supposed to. It seems this is caused by having 2 scopes in keycloak: `manage` and `manager-members`. We may have differentiated this at one point, but now we do now, there are PIs, managers, and members.

This introduces a class UserPermissions that we can use to set these, rather than having to parse keycloak scopes everywhere we want to use permissions.

This also fixes a small bug in a migration.

### DIFF
--- a/projects/views.py
+++ b/projects/views.py
@@ -10,14 +10,10 @@ from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.core.mail import send_mail
 from django.core.validators import validate_email
-from django.db import transaction, DataError
+from django.db import DataError, transaction
 from django.forms.models import model_to_dict
-from django.http import (
-    Http404,
-    HttpResponseForbidden,
-    HttpResponseRedirect,
-    JsonResponse,
-)
+from django.http import (Http404, HttpResponseForbidden, HttpResponseRedirect,
+                         JsonResponse)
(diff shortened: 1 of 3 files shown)


projects/views.py
```diff
@@ -58,6 +47,52 @@ class UserNotPIEligible(Exception):
...
+    def get_manager_projects(keycloak_client, username):
+        """Gets project names for all project this user is a manager of
+        """
+        return [
+            project["groupName"]
+            for project in keycloak_client.get_user_roles(username)
+            if UserPermissions._parse_scopes(project["scopes"]).manage
+        ]
+
+
```
msherman64, line 95: this looks like it's centralizing the mapping from keycloak scopes -> portal permissions, rather than a bunch of if-then? scattered around?

msherman64 (approved):
I'm not super familiar with this part of the code, but the changes look sensible to me.

for clarity, it might be better to separate the logic change from the black reformatting, but no big deal 🤷

projects/views.py
Mark-Powers: Yes. Now all of the permissions/scope code is here, and a block of code can check against this class for a portal permission. This should prevent bugs with manually parsing/interacting with keycloak scopes each time we check a user's role.

PR #457, opened 2024-10-02
Dependency upgrades

Mark-Powers:
- Updates python to 3.10
- Adds mysqlclient to container - used by `./manage.py dbshell`
- Removes reversion - does not migrate cleanly (see PR #63...), and hasn't been used since 2019.
- Removes django-impersonate - Unused, and replaced by keycloak impersonate
- Removes mptt - Outdated query optimization, potentially unused
- Updates django to 4.2, and related django packages
- Remove django-classy-tags - Unused
- Remove kombu - Unused celery extension

Our dependencies are not yet updated to django 5. Django 4.2 is supported until April 2026, and python 3.10 until October 2026. Another thing on the radar is mariadb upgrade, but we are good on 10.11 until Feb 2028.


Mark-Powers:
Impersonate made more sense before federated identity, now you can impersonate a user on portal and in horizon via keycloak. It is still supported, but since we no longer need it, and the logs show that no operator has used it since 2019, we should remove it.

PR #467, opened 2024-10-08
Add memcached library

Mark-Powers:
Django's built in module was removed in 4.0, and so this external library must now be included to continue to use memcached.

https://docs.djangoproject.com/en/5.1/topics/cache/#memcached


msherman64 (approved):
👍

PR #476, opened 2024-10-14
Fix help desk button for logged out users.

Mark-Powers:
If a user is authenticated, they are correctly redirected to the "logged in" form, so it's safe to always link to guest form. I

In case users were using this button to go to "My tickets" instead of submitting a new one, it now conditionally updates the menu bar link to handle this.


msherman64 (approved):
Looks good to me, nice!

PR #486, opened 2024-10-29
Add timezone info to outage emails

Mark-Powers:
This reformats the outage email (both reminder and new/update emails) to include timezone in Chicago and UTC time, for example '2015-07-01 15:00 CDT  (20:00 UTC)'.


msherman64 (approved):
lgtm!

PR #488, opened 2024-11-04
Add project member export feature

Mark-Powers:
PIs have requested to be able to export a list of project members. This adds a button that downloads a CSV file with member information.


msherman64 (approved):
catching up on reviews, this seems sane to me.

PR #489, opened 2024-11-04
Normalize institutions

Mark-Powers:
This adds a "task" that we can run interactively for now to tag users with normalized institution data. This allows us to define institutions with aliases in the DB, and the process checks against this, or prompts for a new alias/institution to be created. The admin page allows us to easily see what users are from what institutions. 

This also fixes some encoding issues from our institutions list, which isn't used anywhere else really. That data isn't perfect and is missing the national labs, international universities, so there is quite a bit of manual information that needs to be entered on the first pass.


msherman64 (approved):
Looks good to me, steady progress from json towards db models

PR #498, opened 2024-11-25
Rewrite allocations admin interface to django

Mark-Powers:
The current allocation admin interface is 9+ years old and aging. A scary recent issue is failure to load depending on keycloak/cloudflare's response time, as at the moment the front end app loads all users and projects. Additionally, it is displaying so many allocations that my browser slows down. Rather than a band-aid fix (e.g. pagination), this PR embeds a hacky HTML popup into the Django admin to accomplish the same thing. This allows us to remove the old Javascript code and eventually its outdated dependencies.

A few "readonly" computed Django fields are added to the allocation admin, which display the same information as the existing SPA. The most complicated one of these includes modal forms for submitting allocation decisions. This form submits data to the same backend API that the SPA does.

image

image

image


allocations/views.py
```diff
@@ -184,44 +181,6 @@ def approval(request):
             except ValueError:
                 errors["computeAllocated"] = "Compute Allocated must be a number."
 
-        if data["computeRequested"]:
```
Mark-Powers, line 187: None of this that is removed was updated, so the validation is not needed.

PR #512, opened 2024-12-19
Add details to pub list, and fix perms

Mark-Powers:
Adds review information to publications in the list view. Also now shows publications from all of a user's projects, not just those they have submitted (which was confusing with managers). Also fixes a delete permission issue: we were not checking ownership at all, and this ensures only the owner can delete what they have submitted.

image


projects/pub_views.py
```diff
@@ -69,16 +68,25 @@ def user_publications(request):
...
+            else:
+                pub.delete_pub()
         except Exception:
             logger.exception("Failed removing publication")
             messages.error(
                 request,
                 "An unexpected error occurred while attempting "
                 "to remove this publication. Please try again",
             )
+    mapper = ProjectAllocationMapper(request)
```
Mark-Powers, line 86: This section now queries for publications for all of the user projects, not just publications the user has submitted.

msherman64 (approved):
looks good to me, thanks for calling out the permissions change

PR #519, opened 2025-01-22
Add support for flavor reservation

Mark-Powers:
Add support in usage enforcement to support flavor reservation. This differs from host reservation charging, as with flavor reservation not all of the host will be used, and so we charge proportional to this usage.

### DIFF
--- a/balance_service/enforcement/usage_enforcement.py
+++ b/balance_service/enforcement/usage_enforcement.py
@@ -226,7 +226,7 @@ class UsageEnforcer(object):
                     self._date_from_string(lease["end_date"])
                 ),
                 hourly_cost=self._get_reservation_sus(
-                    reservation["resource_type"], reservation["allocations"]
+                    reservation
                 ),
             )
             new_charge.save()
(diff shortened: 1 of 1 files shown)


allocations/admin.py
```diff
@@ -11,7 +11,16 @@
...
+    fields = [
+        "region_name",
+        "user",
+        "resource_type",
+        "resource_id",
+        "hourly_cost",
+        "start_time",
+        "end_time",
+    ]
+    readonly_fields = ["region_name", "user"]
```
Mark-Powers, line 23: This updated admin page just helps with debugging/checking on charges.

User really needs to be readonly. Otherwise django loads a 10,000+ item dropdown for each charge with all users.

msherman64 (approved):
overall lgtm.

Are there any backwards compatibility concerns, e.g. this should behave exactly the same for sites that are only running blazar for baremetal?

msherman64 (approved):
This seems totally backwards compatible, no concerns on my part.

Mark-Powers:
This is backwards compatible, it just adds custom logic if blazar sends a reservation of type `flavor:instance`

PR #522, opened 2025-01-31
Add admin API for project allocation details

Mark-Powers:
We'd like to use this API in order to check at a Chameleon site if a project has an active allocation. That allows us to clean up resources for projects that have expired.

This new API endpoint is similar to one we have to "project extras", and implements very basic token auth. This endpoint only exposes very basic information as well, some info about the project (in case a site wants to contact a PI), and information about the allocations. This inclues the last date the project had/has an allocation for, if they currently have an active allocation, and if they have a pending allocation. This way, we can adjust our implementation as we see fit around these parameters.

### DIFF
--- a/allocations/urls.py
+++ b/allocations/urls.py
@@ -6,6 +6,7 @@ app_name = "allocations"
 
 urlpatterns = [
     path("", views.index, name="index"),
+    path("api/view/<str:charge_code>/", views.view_project, name="view_project"),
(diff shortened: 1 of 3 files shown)


msherman64 (approved):
Because we're only using this for KVM@TACC right now, I'd say this is fine.

But we'd want to refactor and use e.g. keycloak for auth if we would eventually want to use this at associate sites.

PR #535, opened 2025-03-05
Validate length of artifact uuid before calling API

Mark-Powers:
This ensures that we check for a sane UUID length before calling the trovi API to prevent abuse. We only need to do this when getting the artifact. This also fixes a 500 when a version slug being request doesn't exist, and instead raises 404.

### DIFF
--- a/sharing_portal/views.py
+++ b/sharing_portal/views.py
@@ -105,6 +105,11 @@ class GitUrlParser:
         return parsed_info
 
 
+def check_url_param_length(param, length=36):
+    if len(param > length):
+        raise Http404("That artifact or version does not exist")
+
+
 def with_trovi_token(view_func):
     def _wrapped_view(request, *args, **kwargs):
         if not request.user.is_authenticated:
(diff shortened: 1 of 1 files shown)


Mark-Powers:
Django has path converters which let you set url param types, but it's deprecated in 5.1. The two best alternatives seem to either be validating in the view (as implemented here), or use a regex path (which I'm avoiding).

msherman64:
Does it make sense to check if the parameter is a valid uuid?

e.g. 

```
try:
    val = uuid.UUID(param)
except ValueError:
    raise Http404("That artifact or version does not exist")
```

sharing_portal/views.py
```diff
@@ -105,6 +105,14 @@ def parse(self, url):
         return parsed_info
 
 
+def check_url_param_length(param, length=36):
+    """If the provided paramter is greater than a given length,
```
msherman64, line 109: typo: parameter

Mark-Powers:
I originally didn't do the full UUID validation because some trovi IDs are legacy, and some do not contain "-" between the UUID parts. I see python UUID handles this second case, and our code already covers the legacy case.

PR #538, opened 2025-04-09
Task cleanup

Mark-Powers:
This fixes some general issues with tasks and the pub import tasks.
- Removes unneeded logged in the consistency check. This hasn't been needed for a long time.
- Removes check charge task. Since portal can't connect to the openstack DBs, this does nothing but generate errors. See the commit for more details, but it means that we are slightly overcharging a few users.
- Changes the pub import tasks. I'm not sure why it was working locally for me before, and wasn't running when staged on dev. With these changes, things are working better.

### DIFF
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -478,7 +478,6 @@ def check_keycloak_consistency():
                 keycloak_client.update_project(project, has_active_allocation="false")
 
 
-@task
 def check_charge():
     """
     Check if the charges of the active allocations are in sync
--- a/allocations/tasks.py
+++ b/allocations/tasks.py
@@ -429,9 +429,6 @@ def check_keycloak_consistency():
         for project in Project.objects.filter(~Q(charge_code__in=active_projects))
     )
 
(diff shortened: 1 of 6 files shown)


allocations/tasks.py
```diff
@@ -484,7 +478,6 @@ def check_keycloak_consistency():
                 keycloak_client.update_project(project, has_active_allocation="false")
 
 
-@task
```
msherman64, line 487: what does removing this do/change?

projects/user_publication/publications.py
```diff
@@ -13,22 +13,22 @@
...
 
 
-@shared_task(bind=True)
+@task(bind=True)
 def import_pubs_semantic_scholar_task(self):
     return import_pubs_task(self, "semantic_scholar")
 
 
-@shared_task(bind=True)
+@task(bind=True)
```
msherman64, line 31: This seems fine, but what's the difference between "task" and "shared task" here?

msherman64 (approved):
It seems fine, there's just a couple parts where I'm not super familiar with how it works.

PR #539, opened 2025-04-14
Allocations/charges fixes

Mark-Powers:
This fixes 2 issues with allocations/charges on the project overview pages
- Usage would appear with negative SUs for a specific user's budget if they deleted a pending lease. When a lease is deleted we just update end time of the charge, and we [check for negative values](https://github.com/ChameleonCloud/portal/blob/master/balance_service/utils/su_calculators.py#L9-L27) typicall. However, we didn't go through these functions for user budgets, since there we are using database queries to calculate the user's usage. This is fixed with a `Case` statement in the django db query.
- Users would get "not enough SUs" error or the "SUs almost used" warning email but their dashboard would show a low number of SUs used. That is because we track 2 numbers: SUs used, and SUs encumbered. This updates the project displays to use "total" (used + encumbered) instead of just "used".

### DIFF
--- a/allocations/admin.py
+++ b/allocations/admin.py
@@ -1,4 +1,6 @@
 from django.contrib import admin
+from django.db import models
+from django import forms
 from django.utils.safestring import mark_safe
 from django.urls import reverse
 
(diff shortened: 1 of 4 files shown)


allocations/admin.py
```diff
@@ -11,7 +13,22 @@
...
+        "end_time",
+        "hourly_cost",
+    ]
+
+    formfield_overrides = {
+        models.TextField: {"widget": forms.TextInput(attrs={"size": "10"})},
+    }
+
+    def has_change_permission(self, request, obj):
+        return False
```
msherman64, line 31: how does this section relate to the two mentioned fixes?

msherman64 (approved):
overall lgtm, just unclear on the purpose of the admin panel change.

PR #540, opened 2025-04-14
Fix overview/help desk loading issues

Mark-Powers:
If the helpdesk isn't working, show an error message instead. This only checks on the ticket list pages. The other ticket pages will show the normal error page. 

The main improvement is this fixes the dashboard loading if RT is not configured or is down.


msherman64 (approved):
+1

PR #541, opened 2025-04-15
Wrap all publication stuff in try/except

Mark-Powers:
This code is just too dependent on outside data to be reliable. Lots of random errors pop-up due to data validation in django/mariadb. This wraps everything in a try/except so that if any one publication causes problems, we can ignore it.


msherman64 (approved):
Seems sane to me. A later refactoring might be to extract each "check THING for publication" into a method returning "parsed_publication | None", to clean up the iteration.

PR #543, opened 2025-05-09
Fix config for local dev

Mark-Powers:
This fixes debug logs, `docker-compose` -> `docker compose`, and some default env variables. Also removes phpmyadmin, since I don't think anyone uses it, and referenceapi since that container image can't be pulled.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,10 @@ services:
     command: ["runserver", "0.0.0.0:${PORTAL_PORT}"]
     depends_on:
       - mysql
+  referenceapi:
+    image: ghcr.io/chameleoncloud/reference-api:latest
+    ports:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -21,10 +19,6 @@ services:
     command: ["runserver", "0.0.0.0:${PORTAL_PORT}"]
     depends_on:
       - mysql
-  referenceapi:
-    image: docker.chameleoncloud.org/referenceapi:latest
-    ports:
-      - "127.0.0.1:8891:8000"
```
msherman64, line 27: can we just push the current referenceapi container to ghcr.io?

msherman64 (approved):
this looks good to me, my only thought is that we could push the current referenceapi container up to ghcr so that it's pullable.

PR #552, opened 2025-08-07
Hide detailed fields in the hardware browser filter, add threads slider

pdmars:
I added two main things here to simplify the view while still keeping the details:
- Show/Hide Detailed Fields for various sections (processor, placement, network devices, SSD)
- Changed # Threads simple filter to a slider instead of checkboxes

Screenshot 2025-08-07 at 10 54 14

Expanding the detailed view on processor:
Screenshot 2025-08-07 at 10 54 30

### DIFF
--- a/g5k_discovery/vue/HardwareCatalogueFilters.vue
+++ b/g5k_discovery/vue/HardwareCatalogueFilters.vue
@@ -46,31 +46,53 @@
         </button>
       </div>
     </div>
-
-    <div class="row">
+    <div class="row simple-filter-row">
+      <div class="col col-sm-2 threads-slider">
+        <strong># Threads</strong>
+        <div class="range-slider">
+          <input
+            type="range"
+            :min="minThreads"
+            :max="threads.max"
+            v-model.number="threads.min"
+            class="range-low"
+            @input="onThreadsChange"
+          />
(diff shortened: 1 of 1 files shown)


g5k_discovery/vue/HardwareCatalogueFilters.vue
```diff
@@ -136,13 +417,83 @@
...
+  "Vendor",
+  "Serial"
+];
+
+const NETWORK_DEVICES_DETAIL_FIELDS = [
+  "Model",
+  "Vendor",
+  "Name",
+  "# Active Devices"
+];
```
pdmars, line 495: Do we want to make these configurable?

Mark-Powers (approved):
I think this looks good! I don't see any need to make this more configurable, it seems easy enough to change in the code if that comes up.

g5k_discovery/vue/HardwareCatalogueFilters.vue
msherman64: I like having the "detail fields" grouped up here, but I'm not sure it makes sense to make them configurable until we have refined the data model some more.

msherman64 (approved):

PR #554, opened 2025-08-14
Improve trovi page with paginated API calls

Mark-Powers:
This adds AJAX pagination to trovi. We must still render server side,
since it would take too much development time to properly handle this
in the frontend. We can't do a typical pagination, with page buttons
since the Trovi API doesn't support this, and that would break
searching. So now it makes repeated requests for artifacts, and portal
responds with SSRed artifact cards.


pdmars (approved):
Nice! Great to add this.

msherman64 (approved):

PR #555, opened 2025-08-26
Add max lease duration and lease update enforcement

Mark-Powers:
Adds the built in blazar lease duration and update enforcement filters to our enforcement API via the balance service.

This also adds a very configurable system so that we can define custom values for each enforcement constant depending on user/project/flavor.


balance_service/enforcement/usage_enforcement.py
```diff
@@ -379,7 +410,7 @@ def _date_from_string(self, date_string, date_format="%Y-%m-%d %H:%M:%S"):
             # Blazar introduces a new dateformat
             new_date_format = "%Y-%m-%dT%H:%M:%S"
             date = datetime.datetime.strptime(date_string, new_date_format)
-        return date
+        return date.replace(tzinfo=pytz.UTC)
```
Mark-Powers, line 413: Previously timezones were being ignored here. They didn't matter since they were just storing lease start and end times for charging, never doing any comparison. That isn't the case anymore because of `_check_lease_update_window` so we need to make them timezone aware.

balance_service/enforcement/usage_enforcement.py
```diff
@@ -336,9 +348,24 @@ def stop_charging(self, data):
                     charge.end_time = now
                     charge.save()
 
-    def __get_billrate(self, resource, resource_type=None):
-        su_factor = resource.get("su_factor")
+    def __get_billrate(self, resource, reservation, resource_type=None):
+        # Do not charge for floating IPs
+        if resource_type == "virtual:floatingip":
+            return 0
```
pdmars, line 354: do we not bill for these at all? or just not in this method's calculation?

balance_service/enforcement/usage_enforcement.py
```diff
@@ -336,9 +348,24 @@ def stop_charging(self, data):
...
+        # SU factor is configured per flavor in portal
+        # Note: this doesn't let us customize su_factor per project/user yet
+        if resource_type == "flavor:instance":
+            flavor_id = _get_reservation_flavor_id(reservation)
+            v = ConfigVariable.get_value(
+                "su_factor",
+                flavor_id=flavor_id,
+            )
+            if v is not None:
+                return float(v)
```
pdmars, line 365: what happens in this whole process is we don't have a value for the flavor in question (ie v is None)? what defaults does this revert back to?

msherman64 (approved):
lgtm!

PR #558, opened 2025-09-26
Update node list and node view with a clean simple look

pdmars:
I tried to simplify the look a bit so it was easier on the eyes to find the information users are generally interested in when they look at the page. I'm not entirely sure how successful I was, but here's my current take. Previous views followed by the new look below. 

The two pages I focused on here are when a user clicks "View" after adjusting filters to find matches, and then once they click on an individual resource to look at its details.

Previous node list:
Screenshot 2025-09-26 at 15 50 37

Previous individual node view:
Screenshot 2025-09-26 at 15 50 52

New node list:
Screenshot 2025-09-26 at 16 30 28

New individual node view:
Screenshot 2025-09-26 at 16 29 57

### DIFF
--- a/g5k_discovery/templates/g5k_discovery/node_details.html
+++ b/g5k_discovery/templates/g5k_discovery/node_details.html
@@ -74,18 +74,6 @@
                         </dl>
                     </div>
                 </div>
-                <div class="capability-group">
-                    <h4 class="capability-group-header" data-toggle="collapse" data-target="#os">
-                        <i class="fa fa-caret-right"></i> Operating System
-                    </h4>
-                    <div id="os" class="collapse capability-group-body">
-                        <dl class="dl-horizontal">
-                            <dt>Kernel</dt><dd>{{ node.operating_system.kernel }}</dd>
-                            <dt>Name</dt><dd>{{ node.operating_system.name }}</dd>
-                            <dt>Version</dt><dd>{{ node.operating_system.version }}</dd>
-                        </dl>
-                    </div>
-                </div>
                  <div class="capability-group">
                     <h4 class="capability-group-header" data-toggle="collapse" data-target="#bios">
                         <i class="fa fa-caret-right"></i> Bios


g5k_discovery/templates/g5k_discovery/node_details.html
```diff
@@ -1,214 +1,184 @@
...
+                        <dl class="dl-horizontal">
+                            <dt>GPU</dt><dd>{% if node.gpu and node.gpu.gpu %}Yes{% else %}No{% endif%}</dd>
+                        </dl>
+                    </div>
                 </div>
-                <div class="col col-sm-3">
-                    <strong>Model: </strong> {{ sd.model }}
+                <div class="capability-group">
+                    <h4 class="capability-group-header" data-toggle="collapse" data-target="#os">
+                        <i class="fa fa-caret-right"></i> Operating System
```
Mark-Powers, line 79: Can we remove this section?

Mark-Powers (approved):
Looks good! I think it is much cleaner now

g5k_discovery/templates/g5k_discovery/node_details.html
```diff
@@ -1,214 +1,184 @@
...
+                    <h4 class="capability-group-header" data-toggle="collapse" data-target="#jobs">
+                        <i class="fa fa-caret-right"></i> Supported Job Types
+                    </h4>
+                    <div id="jobs" class="collapse capability-group-body">
+                        <dl class="dl-horizontal">
+                            <dt>Best Effort</dt><dd>{% if node.supported_job_types.besteffort %}Yes{% else %}No{% endif%}</dd>
+                            <dt>Deploy</dt><dd>{% if node.supported_job_types.deploy %}Yes{% else %}No{% endif%}</dd>
+                            <dt>Virtual</dt><dd>{{ node.supported_job_types.virtual }}</dd>
+                        </dl>
+                    </div>
```
msherman64, line 147: I think "job type" is also unused in chameleon, although I guess we might eventually want to add (and distinguish) containers / vm / baremetal somewhere

msherman64 (approved):
Looks good to me! All I can think of is that it might be nice to have "node_name" somewhere prominent in both views.

pdmars:
> Looks good to me! All I can think of is that it might be nice to have "node_name" somewhere prominent in both views.

I think the node name will be instead of the uuid if it exists on the list page. And then in () on the node view page after the uuid, similar to how it is now. It's just this test setup doesn't have that data for what I've showed here, but it's still in there if it's in the data.

g5k_discovery/templates/g5k_discovery/node_details.html
pdmars: I'm good either way, we can drop it if we want or leave it and eventually repurpose it.

PR #559, opened 2025-09-30
Add redirects to new trovi dashboard, update daypass share form

Mark-Powers:
Redirects most trovi pages to their new dashboard counterparts. I left the original code untouched, but assuming users don't raise issues with the new dashboard, we can rip it out and all of the templates.


pdmars (approved):
Looks good!

PR #588, opened 2026-04-21
Update KS user checking to use username and email

Mark-Powers:
Instead of just looking for a portal user by keystone username, also look up the email.

The long term fix is to store the KC subject ID in the KS user for lookup, which `get_user_by_reference` can already utilize.

I will ensure this works on dev before deploying.

### DIFF
--- a/balance_service/enforcement/usage_enforcement.py
+++ b/balance_service/enforcement/usage_enforcement.py
@@ -28,6 +28,7 @@ from balance_service.enforcement import exceptions
 from balance_service.models import ConfigVariable
 from balance_service.utils import su_calculators
 from projects.models import Project
+from projects.util import get_user_by_reference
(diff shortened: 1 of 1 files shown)


msherman64 (approved):
Logic looks good to me.

Restating for my own understanding:

1. context user_id is captured, and === a keystone userid
2. get the keystone user from keystone user id
3. get user name and user email from keystone user
4. find portal user matching keystone username or email


for right now, this seems totally reasonable.

Longer term, code "shape" wise, it seems like we should have a helper method to do 
"map keystone user ID to portal user"
it wasn't clear to me if the username or email from _get_keystone_user were used anywhere else, and encapsulating this in a method would make the logic linking the "get_user_by_reference" and "get info from keystone" together explicit.

total nit though, and we're gonna change thing anway.

